PR #189, opened 2023-11-04
Patch atomic jack water shader change

1954764916:
I have made some optimizations to the scene to ensure that it runs smoothly on my computer, but it is not clear whether other people's computers run smoothly.

The character control system has not been modified yet, and the native controller is not so smooth to use.


h0lybyte:
Try to break this down into smaller changes instead of a large one and instead of the `main` branch, please pick the `dev` branch.

Issue #163, opened 2023-11-01
[App] : [Saber] : Audio Busses


taylorbrook:
**Core Concept/Theory**
I propose a shared audio bus system to use throughout the project:
- UI
- SFX
- Dialogue
- Music

All feeding into the master bus and set by default to -3db

Other busses may be needed for specific in-game audio effects processing, but they could all feed into one of these busses above and their levels made accessible to the player in the menu.

* * *

pablowunderlich:
Hello, helping out with the audio, be it SFX, music, or implementation.

pablowunderlich:
I have scripted audio systems into Unity, spawning objects and sounds and managing music. Also created a footsteps manager. Could record sounds, create music, or just do implementation

wizcats:
Hello, I'm here to help with the music (:

pablowunderlich:
That sounds great. Let's coordinate. I'm ok for you to take on the music. I can do SFX and implementaion

h0lybyte:
There are some limitations to the Audio within the scope of WebGL.
The [main documentation](https://docs.unity3d.com/2022.3/Documentation/Manual/webgl-audio.html) states that it **does NOT support FMOD.** :C

1954764916:
I hope to use wav format for the audio format. If MP3 is used, compression will be lost. Audio in wav format takes up more memory.

taylorbrook:
my preference is mono wav for sfx and high quality stereo ogg for music. wav for music will make for too big of file sizes, I fear, and ogg can still loop seamlessly.

1954764916:
> my preference is mono wav for sfx and high quality stereo ogg for music. wav for music will make for too big of file sizes, I fear, and ogg can still loop seamlessly.


It is best not to use Ogg format, wav format has the best sound effect

taylorbrook:
@h0lybyte - any opinion on format? see discussion of wav vs mp3 vs ogg above

pablowunderlich:
> @h0lybyte - any opinion on format? see discussion of wav vs mp3 vs ogg above

I recall h0lybyte mentioned OGG since it's a WebGL build. Wav might be too heavy.

taylorbrook:
"How many audio folks are we? How do y'all wanna split the work?"
[...]

PR #204, opened 2023-11-04
#163

pablowunderlich:
Added audio folder and internal structures to start crafting an ambiance


pablowunderlich:
Audio folders added

h0lybyte:
This one is incorrect because its going from patch to main, skipping dev.

PR #220, opened 2023-11-05
Add Interactable and PlayerInteraction Scripts

sjuhyeon:
1. **Interactable.cs**:
   - This script is attached to interactive objects in the game world.
   - It defines interaction behavior for these objects, allowing them to respond to player actions.
   - Provides a way to display prompt messages to the player when they approach interactable objects.

2. **PlayerInteraction.cs**:
   - This script is attached to the player character.
   - It casts a ray to detect nearby interactable objects within a specified distance.
   - When an interactable object is detected and the player triggers the interaction key, the object's interaction behavior is invoked.
   - Additionally, it offers debugging functionality to visualize the ray and the potential to update the UI with prompt messages.

### DIFF
--- a/apps/saber/Assets/Scripts/Interactions/PlayerInteract.cs
+++ b/apps/saber/Assets/Scripts/Interactions/PlayerInteract.cs
@@ -1,0 +1,63 @@
+using UnityEngine;
+
+public class PlayerInteract : MonoBehaviour
+{
+    [Tooltip("The Transform representing the player's position.")]
+    [SerializeField] Transform playerPos;
+
+    [Tooltip("Maximum distance for object interaction")]
+    [SerializeField] float interactionDistance = 3f;
+
+    [Tooltip("Layer(s) for interactable objects")]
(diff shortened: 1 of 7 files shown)


apps/saber/Assets/Scripts/Interactions/PlayerInteract.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        // Generate the interaction ray and display it for debugging.
+        GenerateRay();
+        DisplayRayOnDebug(onDebug);
+    }
+
+    void GenerateRay()
+    {
+        // Create a ray from the player's position in the direction they are facing.
+        ray = new Ray(playerPos.transform.position, playerPos.transform.forward);
```
SpudTatterson, line 31: i think the ray should be shot from the camera (you can use the ScreenPointToRay function) so the player can more easily interact by looking at an object and not turning the character towards it

SpudTatterson (approved):

PR #253, opened 2023-11-09
[Scene2] Aquatic Rampart

dummbyy:


sjuhyeon:
Cosmic lines are interactable and they're being destroyed when player interact with. Any idea about how to check current active line renderers?

```cs
using UnityEngine;

public class CosmicLine : Interactable
{

    protected override void Interact()
    {
        this.gameObject.GetComponentInParent<LineRenderer>().enabled = false;
    }
}
```

sjuhyeon:
> > Cosmic lines are interactable and they're being destroyed when player interact with. Any idea about how to check current active line renderers?
> > ```cs
> > using UnityEngine;
> > 
> > public class CosmicLine : Interactable
> > {
> > 
> >     protected override void Interact()
> >     {
> >         this.gameObject.GetComponentInParent<LineRenderer>().enabled = false;
> >     }
> > }
> > ```
> 
> maybe have some kind of manager that keeps track of the number of lines and you subtract every time you interact or a list of lineRenderers that you remove the lineRenderer from when its interacted

```cs
   // KeyTower.css
    public static int activeLines = 0;
```

PR #413, opened 2023-12-20
Initial revision for safety and bit performance

artech-git:
Remove usage of `.len()` to ensure any utf-8 encoding doesn't result in errors

Port to OnceCell for improved performance and safety

use tokio::Time::Instant for to overcome halted future during execution


h0lybyte (approved):
Okay!

Issue #1209, opened 2024-03-27
https://kbve.com/arcade/ doesn't show the arcade games


lancilloties:
**Describe the bug**
The https://kbve.com/arcade/ page doesn't show the available arcade games KBVE has made.

* * *

**To Reproduce**

1. Go to https://kbve.com/arcade/
2. See error

* * *

**Expected behavior**
A bunch of links/buttons to the arcade games that were made.

* * *

**Screenshots**
![image](https://github.com/KBVE/kbve/assets/160694112/51eec2bb-1020-422b-bb05-630febd6bc62)

* * *

**Desktop (please complete the following information):**

- OS: ubuntu
- Browser: firefox
- Version: latest

* * *

h0lybyte:
Here is the current Zod:

```js

const arcade = defineCollection({
  schema: z.object({
    title: z.string(),
    status: z.boolean().optional(),
    description: z.string(),
    tags: z.array(z.string()), 
    footnote: z.string().optional(),
    author: z.string().default('KBVE Team'),
    img: z.string().default(''),
    date: z.string().optional(),
    url: z.string().optional(),
    icon: z.string().default('https://kbve.com/favicon.svg'),
    swf: z.string().optional(),
    
  }),
});

```

I will go ahead and see what I can do to fix that today, it should not be a tough fix either.

h0lybyte:
I might as well expand out the arcade options, I will include:

"featured" -> If the game should be in the bento grid as a featured game.
"draft" -> Skip the file ? 
"house" -> This is a house game that we built. 

---

Hmm what else could we add ?

lancilloties:
@h0lybyte Perhaps a "promoted" section with third party games which pay for a spot on our website would work? We don't run their game, but we do host a button/link that redirects to their video game.

h0lybyte:
Creating another comment here to test case the n8n too

h0lybyte:
Another comment here?

lancilloties:
> I also realized that the 404 page does not work as expected too.

such memes

lancilloties:
Issue resolved.

PR #1325, opened 2024-04-16
Atomic patch for fishchip arcade button

lancilloties:
Created fishchip arcade menu button: https://github.com/KBVE/kbve/issues/1209

Please review @h0lybyte.


h0lybyte (approved):
+1 I will test case it.

PR #2403, opened 2024-07-14
Service Benefit Block Responsiveness

dladeira:


h0lybyte:
Why not use tailwindcss for the responsive?

dladeira:
It's easier and more customizable with native

Issue #2212, opened 2024-06-29
[Concept] : Improved Register / Onboard Page


h0lybyte:
**Core Concept/Theory**
A clear and concise description of what the concept is. Ex. It would be cool if [...]

The registration/onboarding page was redesigned to be more intuitive, engaging, and user-friendly, reducing friction and improving user retention. This could include elements like a progress bar, interactive tutorials, and personalized prompts.

* * *

**Alternative Ideas**
Is there any other way this concept could be used?

* Implement gamification elements, such as achievements or rewards, to motivate users. 
* Use AI to offer personalized onboarding experiences based on user preferences. 
* Integrate social media sign-up options for easier registration.”

* * *

**Alternative Examples/Sources**
Are there any other references that you can provide?

* Look at platforms like Discord, Duolingo or Slack for examples of engaging onboarding processes.
* Research best practices from UX/UI design blogs or case studies on effective onboarding strategies

* * *

**Additional information**
Add any other context or examples of this concept here.

Consider A/B testing different versions of the onboarding page to measure user engagement. Gathering user feedback through surveys can also provide insights into pain points and areas for improvement. Including animations or visuals can enhance the overall user experience.

h0lybyte:
Oh I made the ticket twice, the other ticket was : https://github.com/KBVE/kbve/issues/2209 
This will be the official ticket for register / login aka onboarding.

h0lybyte:
Register Page will be https://kbve.com/register/ , so its located here : https://github.com/KBVE/kbve/tree/main/apps/kbve.com/src/components/ui/forms

Its currently in svelte, but we could do it another language? Hmm, I think keeping it in svelte will be better.
Do you want me to move the register / login into two new folders in the Astropad, so its easier and cleaner for us to use?

h0lybyte:
The login form is here ->

[packages/astropad/src/login/kbve/LoginForm.svelte](https://github.com/KBVE/kbve/pull/2464/files#diff-df82638b442705989b1b0740b344beb7d29681dc84e81f33aba92b4af6cca04e)

h0lybyte:
The register and onboard page is almsot done, just needs:

- agreement to terms / eula / legal.
- username check client side
- better error management
- optional: visual animation for errors

Will keep this ticket open once I get a chance to wrap back around to updating the register.

h0lybyte:
TODO: Moving the username out of the registration process and then into the onboarding function.

h0lybyte:
[...]

PR #2481, opened 2024-07-26
First version of register authentication portal

dladeira:


h0lybyte:
Just to make it easier to get back to this branch, going to link it with the issue ticket.

https://github.com/KBVE/kbve/issues/2212

dladeira:
> Can you update it so that it uses tailwindcss instead of inline style?

Added a couple colors to the tailwind configuration. Also I'd suggest to keep the height for the image in the stylesheet as for some reason tailwind doesn't have 30rem height, only 24, and that looks rather off.

h0lybyte:
> > Can you update it so that it uses tailwindcss instead of inline style?
> 
> Added a couple colors to the tailwind configuration. Also I'd suggest to keep the height for the image in the stylesheet as for some reason tailwind doesn't have 30rem height, only 24, and that looks rather off.

https://tailwindcss.com/docs/height#arbitrary-values

To do 30rem height in tailwindcss, you would do `h-[30rem]`

dladeira:
The more you know

h0lybyte:
> The more you know

Before you work on login page , do you want me to merge in register ?

h0lybyte (approved):

h0lybyte:
Going to merge this up and I will review it!

PR #5840, opened 2025-08-09
🥜

Fudster:


JStevenP (commented):
Sufficient for review at this time, although the viscosity factor may be required as part of future implementation. Specification is recommended even while not fully required at this time.

h0lybyte:
Requires more nuts and peas, try it again.

PR #10049, opened 2026-04-12
feat(mc): PaperMC lobby + custom Velocity proxy plugin

Fudster:
## Summary
- New PaperMC lobby (`apps/mc/lobby`) — superflat peaceful, LuckPerms + EssentialsX + WorldEdit + AdvancedPortals, joins the existing mc stack as a persistent singleton Deployment in `arc-runners`
- New custom Velocity proxy plugin (`apps/mc/velocity`) — Kotlin, exposes lobby/server-switch commands, replaces the off-the-shelf velocity plugin used previously
- CI manifest entries (`mc_lobby`, `mc_velocity`) so both images publish through the existing ci-docker → utils-publish-docker-image pipeline
- Project files nested under `apps/mc/` to keep the whole MC stack co-located
- EssentialsX + AntiBuild pinned to 2.21.2 (Modrinth IDs `Oa9ZDzZq` / `gjXSrVNQ`) so the bundle installs on Paper 1.21.11 even though Modrinth declares 1.21.8 as max compat — Bukkit BlockBreak/BlockPlace/PlayerInteract APIs are stable across the 1.21.x line and the plugins enable cleanly

## Test plan
- [x] `docker build -f apps/mc/lobby/Dockerfile -t kbve/mc-lobby:latest .` succeeds
- [x] Smoke container reaches `RCON running` and `Done (~20s)`; `rcon plugins` returns all 5: AdvancedPortals, Essentials, EssentialsAntiBuild, LuckPerms, WorldEdit
- [ ] Velocity proxy plugin builds via gradle and the resulting jar loads on velocity:latest
- [ ] After merge, post-publish rewrites `lobby-deployment.yaml` image tag and the cluster picks up the new `kbve/mc-lobby` image
- [ ] Lobby ↔ Fabric backend cross-server transfer works through the new Velocity commands

## Known follow-ups (not in this PR)
- `apps/mc/lobby/` has no `e2e/*.spec.ts` suite yet — `nx run mc-lobby:e2e:ci` boots the container fine but vitest exits with `No test files found`. Needs a real RCON test suite written before the manifest entry can flip to `has_test: true`.
- `mc_velocity` manifest entry needs `"has_test": false` for the same reason.
- MDX versions for both `mc-lobby` and `mc-velocity` need to be bumped above their `version.toml` (both currently `1.0.0`) before the CI version gate will dispatch the first publish.


Fudster:
Closing as duplicate — #10037 already landed the scaffold and #10047 already renamed the slugs. Replacing with a tiny atomic PR that adds the EssentialsX version-pin (`essentialsx:Oa9ZDzZq`, `essentialsx-antibuild:gjXSrVNQ`), since #10047's slug-only fix will fail at the next main-push e2e — Modrinth still has no Paper 1.21.11 builds for either project, max compat is 1.21.8.

PR #10784, opened 2026-05-10
feat(auth,irc-gateway): provider username strict + GoTrue access-token hook

h0lybyte:
## Summary

Two follow-ups stacked on top of the merged JWT \`aud\` fix (#10749, #10751). Both target the IRC username problem where browsers connected as junk nicks like \`adminkbvecom\` derived from email.

### 1. irc-gateway: provider username strict mode

- Drop email/sub fallbacks. Nick is sourced only from Supabase metadata, in priority: \`kbve_username\` → \`user_metadata.preferred_username\` → \`user_metadata.user_name\` → \`user_metadata.nickname\`.
- \`Claims::irc_nick()\` now returns \`Option<String>\`. When \`None\`, both /ws and the IRC TCP gateway refuse the connection — /ws returns 403 with a human-readable reason; IRC sends a NOTICE then closes.
- Sanitizer keeps a-z, A-Z, 0-9, '_', '-', truncates to 16 chars, and prefixes leading digits with '_'.

### 2. GoTrue Custom Access Token hook (\`profile.username\` → JWT)

New migration: \`packages/data/sql/dbmate/migrations/20260510210000_profile_custom_access_token_hook.sql\`

- \`public.custom_access_token_hook(event jsonb) returns jsonb\` — \`SECURITY DEFINER\`, owned by \`supabase_auth_admin\`, locked search_path. Reads \`profile.username\` by \`event->>'user_id'\`, writes \`kbve_username\` into the JWT claims, returns the mutated event.
- Fail-soft on malformed user_id (\`invalid_text_representation\` → return event untouched).
- Strips a stale \`kbve_username\` when the user has no profile row, so claims always reflect DB state.
- Grants USAGE on \`public\` + \`profile\` to \`supabase_auth_admin\`, plus SELECT on \`profile.username\` and a permissive RLS policy (table is otherwise service_role only).
- Verification block asserts every required privilege exists and that anon/authenticated/public can't execute the hook.
- Down migration drops the function, policy, table/schema grants on \`profile\`, but deliberately keeps USAGE on \`public\` — that grant is shared infra this migration does not own.

GoTrue manifest (\`apps/kube/auth/manifests/supabase-gotrue-auth.yaml\`):
\`\`\`yaml
- name: GOTRUE_HOOK_CUSTOM_ACCESS_TOKEN_ENABLED
  value: 'true'
- name: GOTRUE_HOOK_CUSTOM_ACCESS_TOKEN_URI
  value: 'pg-functions://postgres/public/custom_access_token_hook'
\`\`\`

irc-gateway: adds top-level \`Claims.kbve_username\` and prepends it to the \`irc_nick()\` priority. Two new tests cover \"kbve_username wins\" and \"empty kbve_username falls through\".

## Rollout order

1. Apply migration via dbmate (function exists, hook still off — harmless).
2. Apply GoTrue manifest. Pod restart picks up the URI; new JWTs carry \`kbve_username\`.
3. Merge this PR. CI/CD rebuilds irc-gateway image; gateway prefers \`kbve_username\`.

Existing JWTs predate the hook → fall through to user_metadata until next refresh (default lifetime ~1h).

## Test plan

- [x] \`cargo test --bin irc-gateway auth::jwt\` — 26 passed locally
- [x] \`cargo build --bin irc-gateway\` clean
- [ ] Apply migration to supabase DB, verify \`SELECT public.custom_access_token_hook('{...}'::jsonb);\` returns expected event
- [ ] Roll GoTrue, sign in fresh, decode access token, confirm \`kbve_username\`
- [ ] Roll irc-gateway image, connect to chat.kbve.com/chat, confirm nick = canonical username
- [ ] User without \`profile.username\` row: confirm 403 + reason on /ws (not 1006)

### DIFF
--- a/apps/irc/irc-gateway/src/auth/jwt.rs
+++ b/apps/irc/irc-gateway/src/auth/jwt.rs
@@ -32,6 +32,8 @@ pub struct Claims {
     #[serde(default)]
     pub aud: Option<String>,
     #[serde(default)]
+    pub kbve_username: Option<String>,
+    #[serde(default)]
     pub user_metadata: UserMetadata,
     pub exp: u64,
     pub iat: u64,
@@ -40,6 +42,7 @@ pub struct Claims {
 impl Claims {
     pub fn irc_nick(&self) -> Option<String> {
         let candidates = [
+            self.kbve_username.as_deref(),
             self.user_metadata.preferred_username.as_deref(),
             self.user_metadata.user_name.as_deref(),
             self.user_metadata.nickname.as_deref(),
@@ -148,6 +151,7 @@ mod tests {
             email: Some(format!("{sub}@example.com")),
             role: Some("authenticated".to_string()),
             aud: aud.map(String::from),
+            kbve_username: None,
             user_metadata: UserMetadata::default(),
             exp,
             iat,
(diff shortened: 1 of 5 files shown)


h0lybyte:
Superseded — resubmitted as the rebased state of #10781 (same branch atom-05100033-irc-gateway-jwt-aud-fix).

PR #11367, opened 2026-05-25
fix(bevy_chat): WASM client honors skip_registration

h0lybyte:
## Summary
- WASM `ChatClient` always sent `PASS`/`NICK`/`USER` after WS upgrade, duplicating the gateway's pre-registration from the JWT-authenticated handshake. Ergo rejected the second registration and reset the connection.
- Mirror the native client: skip registration when `IrcConfig.skip_registration` is set; still send `JOIN`s.
- Unblocks the browser arcade chat (https://kbve.com/arcade/isometric/) so messages fan out through #general the same way as Tauri native and the chat.kbve.com web embed.

## Test plan
- [ ] Build WASM via CI Bevy workflow + deploy to astro-kbve/public/isometric
- [ ] Open https://kbve.com/arcade/isometric/, sign in
- [ ] Press T to open chat — confirm panel renders
- [ ] Send a message; verify it appears in Discord #general relay and in chat.kbve.com web users
- [ ] Send a message from Discord; verify it arrives in the in-game chat log


h0lybyte:
Superseded — commit already landed on dev as a5b128125 via another path. Closing.

Issue #11362, opened 2026-05-25
feat(astro-kbve,guild-vault): /dashboard/agents — multi-tenant bot management surface


h0lybyte:
## Goal

Self-service dashboard at `/dashboard/agents` that lets any Discord guild owner manage their KBVE-side bot integrations (PATs, webhook secrets, repo allowlists, queue health) without curl/CLI. Reuses the existing `guild-vault` edge function + Supabase Discord OAuth identity (no new auth stack). Unlocks the multi-tenant pivot started in #11262 + #11349 + #11354 + #11360.

## Why now

After #11360 lands, every guild can host its own `gh-webhook` + `gh-backfill` keyed by `github_pat:<guild>` + `github_webhook:<guild>` in Vault. Provisioning today still requires curl to `guild-vault/tokens.set_token` with a hand-crafted Discord provider_token. That's a non-starter for anyone outside h0lybyte. A UI closes the loop.

## Existing primitives

| Piece | Path | What it gives us |
|---|---|---|
| `guild-vault` edge fn | `apps/kbve/edge/functions/guild-vault/` | `tokens.set_token` / `list_tokens` / `delete_token` / `toggle_token`, all gated by `verifyGuildOwnership` (Discord API ownership check, 5-min cache). Decrypted token values never leave service_role. |
| Supabase Discord identity | `apps/kbve/astro-kbve/src/components/auth/AuthBridge.ts` | Signed-in user has Discord `provider_id` + access to `provider_token`. |
| Dashboard pattern | `src/content/docs/dashboard/<x>/index.mdx` + `Astro<X>Dashboard.astro` + React island | Existing for edge / clickhouse / grafana / vm / argo / forgejo etc. Same shell scales to agents. |
| Profile URL anchor | `kbve.com/@<username>` (see [`project_kbve_profile_url`](packages/data/sql/dbmate/migrations/20260510210000_profile_custom_access_token_hook.sql)) | "Your agents" naturally pivots off the same identity surface. |

### P0 — Catalog page

- `src/content/docs/dashboard/agents/index.mdx` + `AstroAgentsDashboard.astro` shell.
- Static cards for known bot integrations (`discordsh-bot` for now; `irc-gateway` + future bots later). Each card shows: name, description, supported services (e.g. `github`, `github_webhook`, `irc`), link to docs, "Manage" CTA → `/dashboard/agents/<bot_key>`.
- No auth required, no IO. Discovery surface.

### P1 — Guild picker + token list (per bot)

[...]

PR #11488, opened 2026-05-31
feat(auth): embed Discord owned_guilds in JWT (P5.8a of #11362)

h0lybyte:
## Summary
- Extends `public.custom_access_token_hook` to embed an `owned_guilds` JWT claim sourced from `auth.identities.identity_data->'owned_guilds'` (the user's Discord identity row)
- Step 1 of 3 in the fix for the "Discord session expired" dead-end on \`/dashboard/agents/*\` — see [#11362](https://github.com/KBVE/kbve/issues/11362) and discussion on PR #11434
- Migration only — no edge fn / frontend changes here

## Why
Today's failure path: the agents dashboard calls Discord's \`/users/@me/guilds\` directly from the browser using \`provider_token\`. When that token ages out (~7 days) the browser gets a 401 and the page blanks with "Discord session expired", even though the Supabase \`access_token\` is still valid for every vault operation downstream.

The fix is to stop relying on a live Discord call to gate page rendering. Instead, the JWT carries the owned-guild list directly. The dashboard reads from JWT claims, never blanks the page, and only prompts re-auth when the user attempts a mutation that needs fresh proof of ownership.

This PR delivers the JWT plumbing. P5.8b adds the bootstrap edge fn that populates \`identity_data\`. P5.8c wires the frontend to read the claim.

## What the hook does now

\`\`\`
event.user_id ──┐
                ├──► profile.username    ──► kbve_username  (existing)
                └──► auth.identities WHERE provider='discord'
                       └─► identity_data->'owned_guilds'
                             └─► filter ^\\d{17,20}$ + LIMIT 50  ──► owned_guilds
\`\`\`

- **Synchronous, no external calls** — JWT mint stays fast. No \`http\`, \`pg_net\`, or \`plpython3u\` extensions involved.
- **Soft-failure** — missing identity / missing column / wrong jsonb type → \`owned_guilds: []\`. Browser treats empty as "not bootstrapped yet" and calls the P5.8b edge fn before flipping to authenticated state.
- **Cap 50** — Discord allows up to 100 owned guilds for phone-verified accounts; the rare 50+ case falls through to the live API path on the edge fn side.

## Trap dodged
The first draft used \`SELECT jsonb_array_elements_text(...) FROM auth.identities ... LIMIT 1\` directly. Combined with the set-returning function, \`LIMIT 1\` collapsed both "pick latest identity row" and "first element of the unnested array" into one truncation → only the first guild ever made it through. Caught during development; smoke check #2 now explicitly asserts \`length == 3\` for a 3-valid + 1-malformed test fixture so this can't regress silently.

## Embedded smoke checks
Both run inside the migration body and \`RAISE EXCEPTION\` (aborting the migration) on failure:

1. **Empty Discord identity** → \`owned_guilds: []\` + existing claims preserved
2. **3 valid + 1 malformed snowflake** → exactly the 3 valid ones, malformed filtered. Test fixture is rolled back via a sentinel-exception subtransaction so \`wallet_on_auth_user_created\` and other trigger side effects on \`auth.users\` are also undone atomically.

## Local stub
\`packages/data/sql/dbmate/init/01-auth-stub.sql\` gains a minimal \`auth.identities\` mirror + grants \`supabase_auth_admin\` USAGE on \`auth\` + SELECT on \`auth.identities\`. Lets this migration and \`20260523033932_tracker_find_claim_identity_by_discord_id\` apply against a fresh \`dev-docker-compose.yml\`. Stub-only — production \`auth.identities\` is Supabase-owned.

## Down
Restores the v1 hook body (kbve_username only). Deliberately does **not** drop the function, since GoTrue is configured to call it and a missing function breaks all sign-in.

## Test plan
- [x] \`dbmate up\` → applies clean, embedded smoke checks pass
- [x] \`dbmate rollback\` → restores v1 hook (no \`owned_guilds\` emitted)
- [x] \`dbmate up\` again → re-applies clean
- [x] Manual probe — synthetic user with no Discord identity → \`{"role":"authenticated","owned_guilds":[]}\`
- [x] Manual probe — synthetic Discord identity with 3 valid + 1 malformed → \`["111...","222...","333..."]\` (malformed dropped)
- [x] No leftover \`auth.users\` row after smoke checks (sentinel rollback verified)
- [ ] Prod deploy via \`ci-dbmate-deploy.yml\` — manual approval at \`kilobase-prod\` env gate

## Next in series
- **P5.8b** — \`discord-bootstrap\` edge fn populates \`auth.identities.identity_data->'owned_guilds'\` via Discord API (service-role write)
- **P5.8c** — \`agentsService.ts\` reads JWT claim, drops the browser-side Discord API call, caches in localStorage, never blanks the page

Part of #11362.

### DIFF
--- a/packages/data/sql/dbmate/init/01-auth-stub.sql
+++ b/packages/data/sql/dbmate/init/01-auth-stub.sql
@@ -16,6 +16,27 @@ CREATE TABLE IF NOT EXISTS auth.users (
     created_at TIMESTAMPTZ DEFAULT NOW()
 );
 
+-- Minimal auth.identities mirror (FK target for tracker.find_claim_identity_by_discord_id
+-- and source for the custom_access_token_hook owned_guilds claim). Schema follows
+-- Supabase's auth.identities so migrations that read identity_data->>'...' or
+-- filter on provider work identically against local + prod.
+CREATE TABLE IF NOT EXISTS auth.identities (
+    id            UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    user_id       UUID NOT NULL REFERENCES auth.users(id) ON DELETE CASCADE,
+    provider      TEXT NOT NULL,
+    provider_id   TEXT NOT NULL,
+    identity_data JSONB NOT NULL DEFAULT '{}'::jsonb,
+    created_at    TIMESTAMPTZ DEFAULT NOW(),
+    updated_at    TIMESTAMPTZ DEFAULT NOW(),
+    UNIQUE (provider, provider_id)
+);
+
+-- supabase_auth_admin needs USAGE on auth + SELECT on auth.identities so the
+-- custom_access_token_hook (SECURITY DEFINER owned by that role) can read
+-- identity_data->'owned_guilds' at JWT mint time.
+GRANT USAGE  ON SCHEMA auth            TO supabase_auth_admin;
+GRANT SELECT ON TABLE  auth.identities TO supabase_auth_admin;
+
 -- Stub auth.uid() — reads request.jwt.claims (matches Supabase auth's
 -- behaviour closely enough for proxy-function tests that set the GUC
 -- via `SET LOCAL request.jwt.claims = '{"role":...,"sub":...}'`.
--- a/packages/data/sql/dbmate/init/01-auth-stub.sql
+++ b/packages/data/sql/dbmate/init/01-auth-stub.sql
@@ -16,27 +16,6 @@ CREATE TABLE IF NOT EXISTS auth.users (
     created_at TIMESTAMPTZ DEFAULT NOW()
 );
 
--- Minimal auth.identities mirror (FK target for tracker.find_claim_identity_by_discord_id
--- and source for the custom_access_token_hook owned_guilds claim). Schema follows
--- Supabase's auth.identities so migrations that read identity_data->>'...' or
--- filter on provider work identically against local + prod.
-CREATE TABLE IF NOT EXISTS auth.identities (
-    id            UUID PRIMARY KEY DEFAULT gen_random_uuid(),
-    user_id       UUID NOT NULL REFERENCES auth.users(id) ON DELETE CASCADE,
-    provider      TEXT NOT NULL,
-    provider_id   TEXT NOT NULL,
-    identity_data JSONB NOT NULL DEFAULT '{}'::jsonb,
-    created_at    TIMESTAMPTZ DEFAULT NOW(),
-    updated_at    TIMESTAMPTZ DEFAULT NOW(),
-    UNIQUE (provider, provider_id)
-);
-
--- supabase_auth_admin needs USAGE on auth + SELECT on auth.identities so the
--- custom_access_token_hook (SECURITY DEFINER owned by that role) can read
--- identity_data->'owned_guilds' at JWT mint time.
-GRANT USAGE  ON SCHEMA auth            TO supabase_auth_admin;
-GRANT SELECT ON TABLE  auth.identities TO supabase_auth_admin;
-
 -- Stub auth.uid() — reads request.jwt.claims (matches Supabase auth's
 -- behaviour closely enough for proxy-function tests that set the GUC
 -- via `SET LOCAL request.jwt.claims = '{"role":...,"sub":...}'`.
--- a/packages/data/sql/dbmate/migrations/20260531005655_auth_hook_owned_guilds_claim.sql
+++ b/packages/data/sql/dbmate/migrations/20260531005655_auth_hook_owned_guilds_claim.sql
@@ -124,13 +124,13 @@ REVOKE ALL ON TABLE profile.discord_bootstrap_cache
 
 GRANT USAGE  ON SCHEMA profile                         TO supabase_auth_admin;
 GRANT SELECT ON TABLE  profile.discord_bootstrap_cache TO supabase_auth_admin;
--- Round 9 #2: explicit schema USAGE for service_role. Function
--- + table privileges alone aren't enough — without USAGE on the
--- containing schema, name resolution of profile.* by service_role
--- (RPC EXECUTE, dashboard SELECT peek) fails with "permission
--- denied for schema profile". Earlier migrations may or may not
--- guarantee this; the grant is idempotent and the assertion below
--- catches future revoke drift.
+-- Round 9 #2 (updated round 11 #2): explicit schema USAGE for
+-- service_role. Function privileges alone are not enough — without
+-- USAGE on the containing schema, service_role cannot resolve
+-- profile.* RPC names and gets "permission denied for schema
+-- profile". Earlier migrations may or may not guarantee this; the
+-- grant is idempotent and the assertion below catches future
+-- revoke drift.
 GRANT USAGE ON SCHEMA profile TO service_role;
 -- Round 10 #1: service_role intentionally has NO direct privilege
 -- on the cache table. Earlier rounds granted SELECT for "dashboard
@@ -150,17 +150,15 @@ GRANT USAGE ON SCHEMA profile TO service_role;
 -- off postgres or if a non-superuser caller gains direct DML.
 ALTER TABLE profile.discord_bootstrap_cache ENABLE ROW LEVEL SECURITY;
 
--- Round 6 #7: policy is SELECT-only (was FOR ALL). Matches actual
--- grants and prevents future privilege drift where someone re-adds
--- UPDATE and a stale FOR ALL policy silently permits it.
+-- Round 11 #1: service_role has NO direct table privileges (the
+-- freshness RPC is the only read path), so no service_role policy
+-- is created. Defensively drop any prior policy name in case a
+-- previous deploy of this migration left one behind. The drift
+-- assertion in the privilege block (round 11 #10) RAISEs if any
+-- service_role policy reappears later — pairs with the no-direct-
+-- table-privilege assertions to lock the privilege model in place.
 DROP POLICY IF EXISTS "service_role_full_access" ON profile.discord_bootstrap_cache;
 DROP POLICY IF EXISTS "service_role_select"      ON profile.discord_bootstrap_cache;
-CREATE POLICY "service_role_select"
-    ON profile.discord_bootstrap_cache
-    AS PERMISSIVE
-    FOR SELECT
-    TO service_role
-    USING (true);
 
 DROP POLICY IF EXISTS "supabase_auth_admin_select" ON profile.discord_bootstrap_cache;
 CREATE POLICY "supabase_auth_admin_select"
@@ -203,6 +201,7 @@ DECLARE
     v_canonical        text[];
     v_now              timestamptz := statement_timestamp();
     v_input_count      integer     := cardinality(COALESCE(p_owned_guilds, ARRAY[]::text[]));
+    v_valid_count      integer;
     v_relinked_user_id uuid;
 BEGIN
     -- Round 7 #7: USING ERRCODE = '22023' (invalid_parameter_value)
@@ -272,27 +271,44 @@ BEGIN
     -- Round 10 #4: silent canonicalisation filter is the chosen
     -- policy — Discord's API isn't fully under our control, so
     -- one bad guild id shouldn't break the whole OAuth refresh.
-    -- But we RAISE LOG when filtering so the drop shows up in
-    -- Postgres logs (log_min_messages=log captures it).
-    IF v_input_count <> cardinality(v_canonical) THEN
-        RAISE LOG 'service_upsert_discord_bootstrap_cache: filtered % of % owned_guilds (NULL/non-snowflake/dup/cap) for user=%',
-            v_input_count - cardinality(v_canonical), v_input_count, p_user_id;
+    -- Round 11 #3 + #4: split the log into two signals so we can
+    -- tell "Discord sent us malformed snowflakes" apart from
+    -- "we deduped or capped duplicates". User-id is omitted from
+    -- these logs since counts alone are the actionable signal
+    -- (avoids spraying user identifiers into pg_log unnecessarily).
+    SELECT count(*)::integer
+      INTO v_valid_count
+      FROM unnest(COALESCE(p_owned_guilds, ARRAY[]::text[])) AS e(g)
+     WHERE g IS NOT NULL
+       AND g ~ '^[0-9]{17,20}$';
+
+    IF v_valid_count <> v_input_count THEN
+        RAISE LOG 'service_upsert_discord_bootstrap_cache: filtered % invalid owned_guild entries (of % input)',
+            v_input_count - v_valid_count, v_input_count;
+    END IF;
+
+    IF v_valid_count <> cardinality(v_canonical) THEN
+        RAISE LOG 'service_upsert_discord_bootstrap_cache: deduped/capped % owned_guild entries (% valid -> % stored)',
+            v_valid_count - cardinality(v_canonical), v_valid_count, cardinality(v_canonical);
     END IF;
 
     -- Account relinking: drop stale row carrying same provider_id
     -- under a different user_id so the upsert below can succeed
     -- against the discord_provider_id UNIQUE constraint without
-    -- raising a raw unique_violation. Round 10 #3: capture the
-    -- displaced user_id into v_relinked_user_id + RAISE LOG so
-    -- relinks are observable in Postgres logs.
+    -- raising a raw unique_violation. Round 10 #3 / 11 #3: capture
+    -- the displaced user_id into v_relinked_user_id + RAISE LOG so
+    -- relinks are observable in Postgres logs. Raw provider_id is
+    -- omitted from the log to avoid leaking it into pg_log; the
+    -- before/after user_ids are kept because that pair is the
+    -- actionable info for incident response.
     DELETE FROM profile.discord_bootstrap_cache
      WHERE discord_provider_id = p_discord_provider_id
        AND user_id <> p_user_id
     RETURNING user_id INTO v_relinked_user_id;
 
     IF v_relinked_user_id IS NOT NULL THEN
-        RAISE LOG 'service_upsert_discord_bootstrap_cache: relinked discord_provider_id=% from user=% to user=%',
-            p_discord_provider_id, v_relinked_user_id, p_user_id;
+        RAISE LOG 'service_upsert_discord_bootstrap_cache: relinked discord provider from user=% to user=%',
+            v_relinked_user_id, p_user_id;
     END IF;
 
     -- Round 10 #5: ON CONFLICT (user_id) DO UPDATE intentionally
@@ -346,15 +362,21 @@ LANGUAGE sql
 STABLE
 SECURITY DEFINER
 SET search_path = ''
+ROWS 1
 AS $$
+    -- Round 11 #5: LEFT JOIN over a single-row input so the caller
+    -- always gets exactly one row back — NULL timestamps + 0
+    -- guild_count when no cache exists. Saves the Axum side an
+    -- explicit empty-result-set branch.
     SELECT
         c.refreshed_at,
         c.updated_at,
         c.created_at,
-        cardinality(c.owned_guilds)::integer AS guild_count
-    FROM profile.discord_bootstrap_cache AS c
-    WHERE p_user_id IS NOT NULL
-      AND c.user_id  = p_user_id;
+        COALESCE(cardinality(c.owned_guilds), 0)::integer AS guild_count
+    FROM (SELECT p_user_id AS user_id) AS input
+    LEFT JOIN profile.discord_bootstrap_cache AS c
+      ON c.user_id = input.user_id
+    WHERE input.user_id IS NOT NULL;
 $$;
 
 ALTER FUNCTION profile.service_get_discord_bootstrap_cache_freshness(uuid)
@@ -528,7 +550,8 @@ BEGIN
     IF NOT has_schema_privilege('supabase_auth_admin', 'profile', 'USAGE') THEN
         RAISE EXCEPTION 'supabase_auth_admin must hold USAGE on schema profile.';
     END IF;
-    -- Round 9 #2: service_role schema USAGE for RPC + dashboard SELECT.
+    -- Round 9 #2 (updated round 11 #2): service_role schema USAGE
+    -- for RPC execution.
     IF NOT has_schema_privilege('service_role', 'profile', 'USAGE') THEN
         RAISE EXCEPTION 'service_role must hold USAGE on schema profile for bootstrap cache RPC access.';
     END IF;
@@ -580,6 +603,19 @@ BEGIN
     END IF;
     IF NOT has_function_privilege('service_role', 'profile.service_get_discord_bootstrap_cache_freshness(uuid)', 'EXECUTE') THEN
         RAISE EXCEPTION 'service_role must hold EXECUTE on profile.service_get_discord_bootstrap_cache_freshness for dashboard freshness peeks.';
+    END IF;
+    -- Round 11 #10: drift guard. Any future PR that re-creates a
+    -- service_role RLS policy on this table will fail the
+    -- migration here, forcing the author to explicitly justify
+    -- giving service_role table-level read access.
+    IF EXISTS (
+        SELECT 1
+          FROM pg_policies
+         WHERE schemaname = 'profile'
+           AND tablename  = 'discord_bootstrap_cache'
+           AND roles @> ARRAY['service_role']::name[]
+    ) THEN
+        RAISE EXCEPTION 'no service_role RLS policy may exist on profile.discord_bootstrap_cache; service_role reads go through the freshness RPC.';
     END IF;
     IF has_table_privilege('service_role', 'profile.discord_bootstrap_cache', 'INSERT') THEN
         RAISE EXCEPTION 'service_role must NOT have direct INSERT on profile.discord_bootstrap_cache; use service_upsert RPC.';
--- a/packages/data/sql/dbmate/migrations/20260531005655_auth_hook_owned_guilds_claim.sql
+++ b/packages/data/sql/dbmate/migrations/20260531005655_auth_hook_owned_guilds_claim.sql
@@ -85,7 +85,7 @@ REVOKE EXECUTE ON FUNCTION profile.discord_owned_guilds_are_valid(text[])
 CREATE TABLE profile.discord_bootstrap_cache (
     user_id              uuid PRIMARY KEY,
     discord_provider_id  text NOT NULL
-                              CHECK (discord_provider_id ~ '^[0-9]{15,25}$'),
+                              CHECK (discord_provider_id ~ '^[0-9]{17,20}$'),
     owned_guilds         text[] NOT NULL DEFAULT ARRAY[]::text[],
     refreshed_at         timestamptz NOT NULL DEFAULT NOW(),
     created_at           timestamptz NOT NULL DEFAULT NOW(),
@@ -111,7 +111,7 @@ CREATE TABLE profile.discord_bootstrap_cache (
 );
 
 COMMENT ON TABLE profile.discord_bootstrap_cache IS
-    'Per-user Discord bootstrap cache. Populated exclusively via profile.service_upsert_discord_bootstrap_cache RPC. Read by custom_access_token_hook at JWT mint to embed owned_guilds claim. service_role has SELECT only; INSERT/UPDATE/DELETE flow through the RPC for centralised validation + canonicalisation. supabase_auth_admin has SELECT for the hook.';
+    'Per-user Discord bootstrap cache. Populated exclusively via profile.service_upsert_discord_bootstrap_cache RPC; service_role has SELECT only (no direct INSERT/UPDATE/DELETE) so all writes are funnelled through the RPC for centralised validation, canonicalisation, advisory-lock serialisation, and account-relink handling. SELECT is granted to service_role specifically so dashboard freshness peeks and the RPC''s ON CONFLICT visibility work without going through another RPC layer. supabase_auth_admin has SELECT for custom_access_token_hook at JWT mint.';
 
 REVOKE ALL ON TABLE profile.discord_bootstrap_cache
     FROM PUBLIC, anon, authenticated;
@@ -179,11 +179,14 @@ DECLARE
     v_canonical text[];
     v_now       timestamptz := statement_timestamp();
 BEGIN
+    -- Round 7 #7: USING ERRCODE = '22023' (invalid_parameter_value)
+    -- on all validation failures so service callers (axum-kbve etc.)
+    -- can match on a stable SQLSTATE instead of message text.
     IF p_user_id IS NULL THEN
-        RAISE EXCEPTION 'user_id required';
+        RAISE EXCEPTION 'user_id required' USING ERRCODE = '22023';
     END IF;
-    IF p_discord_provider_id IS NULL OR p_discord_provider_id !~ '^[0-9]{15,25}$' THEN
-        RAISE EXCEPTION 'invalid discord_provider_id';
+    IF p_discord_provider_id IS NULL OR p_discord_provider_id !~ '^[0-9]{17,20}$' THEN
+        RAISE EXCEPTION 'invalid discord_provider_id' USING ERRCODE = '22023';
     END IF;
 
     IF p_refreshed_at IS NULL THEN
@@ -191,17 +194,27 @@ BEGIN
     END IF;
     IF p_refreshed_at > v_now + interval '30 seconds' THEN
         RAISE EXCEPTION 'refreshed_at too far in future (got %, max %)',
-            p_refreshed_at, v_now + interval '30 seconds';
+            p_refreshed_at, v_now + interval '30 seconds'
+            USING ERRCODE = '22023';
     END IF;
     IF p_refreshed_at < '2025-01-01'::timestamptz THEN
         RAISE EXCEPTION 'refreshed_at below floor (got %, min 2025-01-01)',
-            p_refreshed_at;
+            p_refreshed_at
+            USING ERRCODE = '22023';
     END IF;
 
-    -- Round 6 #4 + #5: serialise concurrent writes for this account.
-    -- Always lock user first, then provider, for consistent ordering.
-    PERFORM pg_advisory_xact_lock(hashtextextended(p_user_id::text, 1));
-    PERFORM pg_advisory_xact_lock(hashtextextended(p_discord_provider_id, 2));
+    -- Round 7 #4: single coarse advisory lock for the entire relink
+    -- namespace. The previous per-key locks on (user_id, provider_id)
+    -- closed the same-account race but left a rare cross-swap race
+    -- open: concurrent calls A→Y / B→X each locked their own keys
+    -- but neither held the other side's lock before DELETE'ing the
+    -- target provider row. A single namespace lock serialises ALL
+    -- relinks. The cost is one global serialisation point at human-
+    -- OAuth call rates (≤ hundreds/sec at peak), which is well
+    -- within Postgres advisory-lock contention budgets.
+    PERFORM pg_advisory_xact_lock(
+        hashtextextended('profile.discord_bootstrap_cache.relink', 0)
+    );
 
     -- Canonicalise: NULL + non-snowflake elements filtered; duplicates
     -- collapsed by first-seen order; capped at 50 (matches table CHECK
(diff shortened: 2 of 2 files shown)


h0lybyte:
Folded in 11 of 16 review points (commit c78f22016). Disposition:

**Taken — load-bearing**
- **#2** Moved state to `profile.discord_bootstrap_cache`. Hook reads via PK lookup; no more `auth.identities.identity_data` dependency. This kills #8 + #9 as a side effect (smoke test no longer leans on Supabase's auth.identities shape).

**Taken — hardening**
- **#3** `[0-9]` over `\d` (locale-safe, same cost)
- **#4** `WITH ORDINALITY` + explicit `jsonb_agg(g ORDER BY ord)` (deterministic claim order)
- **#5** `GROUP BY g + MIN(ord)` (dedup, first-seen order preserved)
- **#6** Invalid / missing `user_id` strips `kbve_username` + `owned_guilds` from inbound claims before returning (attacker can't smuggle a privileged claim via a garbage `user_id`)
- **#10 / #16** Typecheck `event->'claims'` as object before `jsonb_set`; non-object input rebuilt as `{}`
- **#11** Soft size cap (`>1200` chars → `[]`), soft degradation rather than failing sign-in
- **#15** CTE rewrite (`src` → `valid` → `jsonb_agg`) adapted to read from the cache table

**Declined with reasoning**
- **#1** Index on `auth.identities` — moot once we stopped reading from there. If we ever do, ships as its own transaction:false migration with explicit operational-risk acceptance.
- **#7** `VOLATILE` over `STABLE` — hook is called once per token mint event, never in set queries. `STABLE` is semantically correct (result is determined by the input + queried rows within a statement). `VOLATILE` adds no safety, just disables a planner optimization that doesn't fire on this code path.
- **#12** `ALTER DEFAULT PRIVILEGES` — cross-cutting, affects unrelated apps. Belongs in its own blast-radius-analysed hardening PR, not this one.
- **#13** Move hook to `auth_hooks` schema — needs a coordinated GoTrue config flip (`GOTRUE_HOOK_CUSTOM_ACCESS_TOKEN_URI`). Separate PR.
- **#14** Split into helper fn — reviewer agrees, inline is fine.

**Smoke checks now cover 5 scenarios** (all run during migration apply, abort migration on failure):
1. Empty event / no user_id → claims preserved, no privileged claims emitted
2. Malformed user_id with smuggled `owned_guilds` claim → attacker claim stripped
3. Non-object claims input → rebuilt as `{}`, `owned_guilds: []`
4. Valid user_id, no cache row → `owned_guilds: []`
5. Valid user_id, cache row with 3 valid + 1 duplicate + 1 malformed → `[valid_a, valid_b, valid_c]` in first-seen order (dedup + filter + LIMIT-1-on-unnest regression guard)

**Local verification**: dbmate up + rollback + up cycle clean. Probes confirmed:
- Real cache row `[777, 666, 777-dup, garbage]` → hook returns `["777","666"]` (dedup + filter + order preserved)
- Malformed user_id with smuggled `"attacker"` claim → hook returns `{"role":"authenticated"}` (attacker claim stripped)
- Post-rollback: cache table dropped, v1 hook returns no `owned_guilds`

h0lybyte:
Second-round review folded in (commit 2be84c7ee). 11 of 15 taken, 3 declined w/ reasoning, 1 no-op.

**Load-bearing behaviour shifts**
- **#1** Dropped smoke Scenario 5 — the only writer to `auth.users` in the migration. Migration is now zero-write to `auth.users`. The LIMIT-1-on-unnest regression guard moves to a dedicated pgTAP suite that runs against the kilobase production image in CI (separate follow-up PR).
- **#10** **7-day freshness window** on the cache row lookup via `statement_timestamp() - interval '7 days'`. Stale rows yield `[]` from the hook so the browser triggers the bootstrap edge fn to refresh instead of trusting stale ownership forever.
- **#11** **STABLE → VOLATILE**. Conceded on second pass — function reads mutable cache state in a login flow; perf cost is theoretical (no set-query memoization opportunity here).

**DDL hardening**
- **#3** Replaced `GRANT ALL TO service_role` with explicit `GRANT SELECT, INSERT, UPDATE, DELETE`. No more TRIGGER/REFERENCES/TRUNCATE for the edge fn.
- **#4** Explicit `REVOKE ALL FROM PUBLIC, anon, authenticated` on the cache table + paired negative assertion in the privilege smoke that fails the migration if anon/authenticated ever gain SELECT.
- **#6** Service-role policy `WITH CHECK` now mirrors the shape constraints — RLS gives clean failure semantics before the CHECK kicks in.
- **#7** New `profile.discord_owned_guilds_are_valid(jsonb)` IMMUTABLE validator + `CHECK` constraint on the cache table. Cache table never stores junk at rest even if the edge fn writes garbage; hook still filters defensively at read (belt + suspender).
- **#14** Named FK constraint `discord_bootstrap_cache_user_id_fkey` (was autogenerated).

**Size + rollback hygiene**
- **#8** Raised soft size cap from 1200 → 1600 chars (50 snowflakes serialize to ~1200 exactly with no margin; 1600 buys ~400 chars headroom).
- **#12** Down migration **preserves the v2 hardening** (object typecheck, invalid-user stripping) instead of restoring the weaker v1 body. Rollback removes the owned_guilds path, not the security posture.
- **#13** Down explicitly strips `owned_guilds` from inbound claims before any other work, so a mid-flight rollback can't leak v2 state into a v1 JWT.

**Declined**
- **#2** `DEFERRABLE INITIALLY IMMEDIATE` FK — single-statement upsert doesn't need it; add when an actual transactional pattern demands it.
- **#5** Drop `supabase_auth_admin_select` RLS policy — with RLS enabled the SELECT GRANT alone is insufficient; without the policy the hook's SELECT would be denied.
- **#9** `refreshed_at` index — no janitor planned yet; add when cleanup logic lands.
- **#15** No-op — reviewer agrees `jsonb_path_query_array` isn't an improvement here.

**Local verification — apply + rollback + apply cycle**

Real-data probes against applied state:

| Probe | Input | Output | Pass |
|---|---|---|---|
| A_fresh | cache row with `[777, 666, 777-dup, 555]` (refreshed_at = NOW) | `["777","666","555"]` | dedup + first-seen order + freshness pass |
| B_stale | same row, `refreshed_at = NOW - 8 days` | `[]` | freshness window correctly gates stale |
| C_attacker | malformed user_id + smuggled `"attacker"` claim | `{"role":"authenticated"}` | attacker stripped |
| D_validator | `INSERT ... owned_guilds = ["garbage"]` | `check_violation` raised | validator rejected bad write at storage layer |

Rollback restores v1 hook with hardening preserved (attacker claim still stripped, non-object claims still rebuilt, inbound `owned_guilds` explicitly dropped, cache table + validator both removed).

**Follow-ups owned outside this PR**
- pgTAP test suite for the LIMIT-1-on-unnest regression (CI-only)
- Switch `dev-docker-compose.yml` from `postgres:17-alpine` to `ghcr.io/kbve/postgres:17.4.1.069-kilobase` so local matches prod ecosystem (incl. real `auth.*` internals + trigger surface)

h0lybyte:
Third-round fold-in pushed (commit 7aa961bdd). Two near-duplicate review blocks consolidated to 12 unique points; 8 taken, 3 declined w/ reasoning, 1 deferred.

**Standout fixes**
- **Validator CASE rewrite + STRICT** — SQL boolean AND doesn't guarantee short-circuit; the old AND chain would let \`jsonb_array_length\` / \`jsonb_array_elements_text\` be planned and evaluated on non-array input and throw. CASE forces order. STRICT documents NULL-input intent.
- **5-minute future-timestamp guard in hook** — a bad write with future \`refreshed_at\` would otherwise keep stale ownership trusted longer than the 7-day window. Probe \`B_future_attack\` confirms a row 10 minutes in the future returns \`[]\`.
- **Validator created before table; \`CREATE TABLE IF NOT EXISTS\` removed** — cleanest order, lets the validator be inlined as a table constraint, and makes schema drift fail loudly instead of silently reusing a stale table.

**Taken (8/12)**
- **#A** Validator CASE + STRICT (correctness short-circuit)
- **#B** REVOKE EXECUTE on validator from PUBLIC/anon/authenticated, GRANT to service_role + supabase_auth_admin
- **#C** Inline 5-minute future-timestamp guard
- **#D** SELECT + UPDATE privilege assertions for service_role (upsert needs both); skipped DELETE assertion since edge fn doesn't currently delete
- **#E** Fixed table comment — dropped misleading "Owned by service_role" wording
- **#F** Validator first → table second; inline constraints; dropped \`CREATE TABLE IF NOT EXISTS\`
- **#I** \`src\` CTE no longer re-validates \`jsonb_typeof\` (table CHECK + validator already guarantee array shape at storage); per-element regex stays as belt+suspender
- **#J** Inlined freshness interval; dropped \`v_freshness\` PL/pgSQL variable

**Declined w/ reasoning (3/12)**
- **#G** Keep all three CHECK constraints. Probe \`D_validator\` proves \`_is_array\` surfaces first, not the generic \`_valid\` catch-all — that's the debuggability win the reviewer explicitly called out as the keep-them tradeoff.
- **#K** No explicit policy drops in down. \`DROP TABLE\` cascades policies — reviewer says "not required".
- **#L** No \`ALTER TABLE OWNER TO service_role\`. Reviewer's own warning (their #11): table owners bypass RLS unless \`FORCE ROW LEVEL SECURITY\` is on. Comment fix (E) addresses the wording instead.

**Deferred (1/12)**
- **#H** \`refreshed_at\` index — no janitor planned yet.

**Local verification — apply + rollback + apply cycle**

| Probe | Input | Output | Pass |
|---|---|---|---|
| A_fresh | cache row `[777, 666, 777-dup, 555]` (refreshed_at = NOW) | `["777","666","555"]` | dedup + first-seen order |
| B_future_attack | row 10 min in future | `[]` | **5-min future guard fires** |
| C_stale | row 8 days old | `[]` | 7-day staleness guard fires |
| D_validator | `INSERT ... owned_guilds = {"not":"array"}` | `check_violation` from `_is_array` constraint | three-constraint debuggability win |

Also 9 inline validator smokes (non-array object, string, number, jsonb null, SQL NULL via STRICT, empty array, valid 2-snowflake, malformed element, too-short snowflake) — all run during migration apply, abort on failure.

**Privilege assertions added** — service_role SELECT+UPDATE on cache table; validator EXECUTE for service_role + supabase_auth_admin; negative assertion that anon/authenticated cannot EXECUTE the validator.

**Follow-ups still owned outside this PR**
- pgTAP regression test for the LIMIT-1-on-unnest pattern (replaces the dropped Scenario 5)
- \`dev-docker-compose.yml\` → \`ghcr.io/kbve/postgres:17.4.1.069-kilobase\` for local-to-prod parity

h0lybyte:
GPT round-4 fold-in pushed (commit c08131325). 7 of 18 taken, 8 declined w/ reasoning, 2 deferred, 1 explicit flip-flop on volatility.

**Standout structural changes**
- **`owned_guilds` storage flipped from `jsonb` to `text[]`** (#1). Discord snowflakes are flat strings — jsonb traversal in the JWT mint hot path was overkill. Validator, hook CTE, and table CHECK now operate on native arrays; only the final claim is wrapped via `to_jsonb(array_agg(...))`. One `unnest` replaces one `jsonb_array_elements_text` per JWT mint.
- **`discord_provider_id` is now UNIQUE** (#4). Mirrors Supabase's own `auth.identities (provider, provider_id)` integrity. Prevents split-brain from edge fn bugs caching the same Discord account under two `user_ids`.
- **Future-timestamp tolerance 5min → 30s** (#8). Old 5min was over-conservative; gave an attacker a 4.5-minute trust extension. 30s covers real Deno-vs-Postgres in-cluster skew.
- **`VOLATILE` → `STABLE`** (#12). Direct flip from round 2 #11 which I'd taken. Round 2 reviewer pushed VOLATILE on pure safety; round 4 reviewer correctly flagged STABLE as the documented semantics for a read-only function. Reviewer-vs-reviewer tiebreak goes to round 4 on technical-correctness grounds.

**Other taken**
- **#9** `timestamps_floor` CHECK rejecting timestamps before 2025-01-01 (combined `refreshed_at` + `created_at` + `updated_at` into one constraint with a single error message)
- **#11** Local `v_now timestamptz := statement_timestamp()` in the hook — safer freshness edits, no repeated function calls
- **#17** `created_at` + `updated_at` columns — `refreshed_at` tracks Discord-data freshness, the new columns track row lifecycle. Edge fn maintains `updated_at` on upsert.

**Declined w/ reasoning (8/18)**
- **#2** Pre-computed claim column — premature optimization, hook CTE is microseconds
- **#5** Simplify RLS WITH CHECK to `true` — round 2 reviewer explicitly asked for shape validation as defense-in-depth (their #6 then); keeping it. Cost is microseconds.
- **#7** FORCE ROW LEVEL SECURITY — reviewer themselves flags "can break assumptions"
- **#10** Numeric range / DOMAIN — DOMAIN worth it when snowflakes appear in 3+ tables (currently 1)
- **#13** Defensive `ORDER BY + LIMIT 1` on username — PK already guarantees uniqueness
- **#14** `ON UPDATE CASCADE` — `auth.users.id` UUID never changes
- **#15** Move all smokes to CI — remaining smokes don't touch `auth.users`; migration time ~30ms
- **#18** Split raw + claim caches — solves a problem we don't have

**Deferred (2/18)**
- **#3** `refreshed_at` index — no janitor yet
- **#6** Write-path RPC wrapper — belongs in P5.8b when the edge fn lands and can choose between direct CRUD vs service RPC

**Local verification — 7 real-data probes all green**

| Probe | Input | Output | Pass |
|---|---|---|---|
| A_fresh | text[] `[777, 666, 777-dup, 555]` (NOW) | `["777","666","555"]` | dedup + first-seen + text[] backend |
| B_future_attack | row 35s in future | `[]` | **30s guard fires, tighter than the old 5min** |
| C_stale | row 8 days old | `[]` | 7-day window unchanged |
| D_UNIQUE | INSERT with colliding `discord_provider_id` | `unique_violation` | new UNIQUE constraint |
| E_validator | INSERT with `ARRAY['garbage']` | `check_violation` from `_owned_guilds_valid` | validator on text[] |
| F_timestamps_floor | INSERT with `refreshed_at='1999-01-01'` | `check_violation` from `_timestamps_floor` | historical-ts attack blocked |
| G_audit_cols | INSERT with defaults | `created_at` + `updated_at` both within 1min of insert | lifecycle audit works |

Rollback drops table + text[] validator overload cleanly; hook v1 hardening preserved per round 2 fold-in.

**Follow-ups still owned outside this PR**
- pgTAP regression test for the LIMIT-1-on-unnest pattern
- `dev-docker-compose.yml` → kilobase image for prod-parity

**Meta note**: 4 review rounds, ~50 total disposition points addressed. Major hardening blocks closed; remaining feedback is largely polish + future-design. Open to one more pass if substantive new concerns surface, otherwise ready for prod deploy via `ci-dbmate-deploy.yml`.

h0lybyte:
Round-5 fold-in pushed (commit 8e60046c9). 11 of 15 taken, 2 declined, 2 deferred.

**Standout — real correctness bug fixed**
- **#5** validator was accepting `ARRAY['valid', NULL]`. Reason: `g !~ '...'` returns NULL for NULL `g`, not true, so `NOT EXISTS` never matched and validator returned true. Fixed by adding `g IS NULL OR` to the WHERE. **New smoke check F asserts this regression can't return.**

**Standout — service RPC write path**
- **#2** service_role no longer has direct INSERT/UPDATE/DELETE on the cache. All writes flow through `profile.service_upsert_discord_bootstrap_cache(...)` RPC, which centralises in one place:
    - Input validation with clean error messages
    - **#7** Canonicalisation: dedup + first-seen ordering + NULL/regex filter + cap 100 (storage now canonical, not just the JWT claim)
    - **#10** Future-timestamp rejection >30s (table CHECK can't use `now()` since it's not IMMUTABLE)
    - **#12** Account relinking: DELETE existing row with same provider_id under different user_id before upsert (avoids raw `unique_violation` leaking to caller)
    - **#1** `updated_at` maintenance on every write (no trigger needed since RPC owns the write surface)
- **Negative privilege assertions added**: migration `RAISE EXCEPTION` if service_role ever regains direct INSERT/UPDATE/DELETE.

**Hook simplified (#3, #8)**
- WITH CHECK → `true` (service_role no longer takes direct write paths)
- Hook drops GROUP BY (storage is canonical) but keeps defensive NULL + regex filter as belt+suspender — JWT mint is a security boundary

**Other taken**: #4 single COALESCE in validator, #6 hook defensive NULL filter, #9 `updated_at >= created_at` CHECK folded into timestamps_ordering_and_floor, #13 down drops the new RPC in proper dependency order.

**Declined**: #14 rename validator (cosmetic; reviewer agrees low priority), #15 DOMAIN for snowflake (premature — one column).

**Deferred**: #11 refreshed_at index (no janitor planned).

**Local verification — 6 real-data probes all green**

| Probe | Input | Output | Pass |
|---|---|---|---|
| A_canonical_stored | RPC input `[777, 666, 777-dup, garbage, NULL, 555]` | storage `{777, 666, 555}` | dedup + NULL filter + garbage filter + first-seen order |
| B_hook_output | hook reads canonical storage | clean JWT claim | hook no longer reruns dedup |
| C_relink | provider_id moves user 66 → user 77 | old row deleted, new row present | account relinking handled |
| D_future_attack | RPC called with `NOW() + 5 minutes` | `service_upsert_...: p_refreshed_at too far in future` | clean error message, not raw constraint violation |
| E_rpc_updated_at | Second upsert via RPC | `updated_at > created_at = t` | RPC maintains updated_at |
| F_validator_null | `ARRAY['valid', NULL]` | `false` | **round 5 #5 NULL bypass fixed** |

Rollback drops table + validator + RPC cleanly in proper dependency order; hook v1 hardening preserved per round-2 fold-in. Re-apply clean.

**5 review rounds done, ~65 disposition points addressed.** The write surface is now what the reviewer kept asking for: one RPC, centralised validation, canonical storage. Hook stays minimal at the JWT mint security boundary.

**Follow-ups still owned outside this PR**
- pgTAP regression test for the LIMIT-1-on-unnest pattern
- `dev-docker-compose.yml` → kilobase image for prod-parity local dev

h0lybyte:
Round-6 fold-in pushed (commit 00da91d3b). 12 of 19 taken, 2 declined, 3 deferred, 2 no-ops on already-correct points.

**Standouts**
- **#1 SQL helper for guild claim** — `profile.get_discord_owned_guilds_claim(uuid, timestamptz)` STABLE + COST 20 does the cache lookup; hook delegates instead of carrying a CTE. Lets the planner inline + cost-estimate.
- **#4 + #5 Advisory locks** in service_upsert RPC on `(user_id, provider_id)` in fixed order — serialises concurrent OAuth refreshes for the same Discord account so DELETE-then-INSERT relinking can't race the UNIQUE constraint.
- **#9 RPC owner pinned to `postgres`** — SECURITY DEFINER behaviour now stable across local/CI/prod instead of varying with the migrating role. Negative assertion in the privilege block verifies.
- **#15 Storage cap 100 → 50** — JWT mint already caps at 50; extra headroom was dead weight. `cardinality()` CHECK replaces the old `COALESCE(array_length, 0)`.
- **#3 Length-check dropped** — unreachable with cap=50.

**Other taken**: #6 validator EXECUTE revoked from all roles (CHECK runs as table owner), #7 RLS policy `service_role_select` is SELECT-only (was `FOR ALL`), #10 RPC error messages trimmed of function-name prefixes, #11 explicit NULL handling for `p_refreshed_at`, #14 `cardinality()` over `COALESCE(array_length)`, #18 `COST 20` on the helper.

**Declined**: #2 pre-computed claim column (third decline; same reason — adds sync invariant, helper makes mint cheap enough), #12 DOMAIN for snowflake (premature, one column).

**Deferred**: #8 FORCE RLS (reviewer's own caveat + postgres-owner bypasses anyway), #13 refreshed_at index (no janitor), #16 relinking smoke (needs `auth.users` write → CI pgTAP).

**No-op**: #17 down doesn't revoke schema USAGE (reviewer notes v1 needs it), #19 validator not SECURITY DEFINER (already INVOKER).

**Local verification — 8 real-data probes**

| Probe | Check | Result |
|---|---|---|
| A_hook_via_helper | hook delegates to helper | JWT claim `["777","666","555"]` |
| B_helper_direct | helper callable standalone | same shape |
| C_stored_cap50 | RPC input of 60 elements | stored as 50 (RPC cap fires) |
| D_validator_cap50 | direct INSERT of 51 | `check_violation` from `_owned_guilds_size_cap` |
| E_validator_grants_revoked | service_role EXECUTE on validator | `f` ✓ |
| F_policy_select_only | RLS policy `polcmd` | `r` (SELECT only) ✓ |
| G_rpc_owner | RPC owner | `postgres` ✓ |
| H_null_refreshed_at_default | NULL `p_refreshed_at` | defaulted to NOW; stored row refreshed within 5s |

All 4 new objects (table, validator, RPC, helper) drop cleanly on rollback in proper dependency order. Re-apply green.

**6 review rounds done, ~85 disposition points addressed total.** Hook is now slim PL/pgSQL orchestration delegating to a planner-friendly SQL helper. Write path is RPC-only with advisory-lock serialisation. Storage cap matches JWT cap. Owner is pinned. Policy matches grants.

**Follow-ups still owned outside this PR**
- pgTAP regression test for the LIMIT-1-on-unnest pattern (CI-only)
- `dev-docker-compose.yml` → `ghcr.io/kbve/postgres:17.4.1.069-kilobase` for prod-parity local dev

h0lybyte:
Round-7 fold-in pushed (commit c2b172420). 8 of 9 taken, 1 deferred.

**Standouts**
- **#4 Cross-swap race closed.** Round-6 per-key advisory locks on `(user_id, provider_id)` serialised same-account writes but left a rare cross-swap open: concurrent A→Y / B→X each held their own per-key locks but neither blocked the other's DELETE-then-INSERT. Replaced both per-key locks with a **single coarse advisory lock on the relink namespace** — serialises ALL relinks. Cost is one global serialisation point at human-OAuth call rates, well within Postgres advisory-lock budgets.
- **Stale comment was a real correctness bug**. Table COMMENT still claimed "service_role has CRUD" — left over from before round-5 dropped service_role to SELECT-only. Updated to reflect actual posture + rationale.

**Other taken**
- **#1** `supabase_auth_admin` SELECT on `profile.username` assertion (catches v1 grant drift that would silently break the kbve_username arm)
- **#2** Explicit `v_claims := v_claims - 'owned_guilds'` before `jsonb_set` for valid users — security boundary visible, matches invalid-user branches
- **#3** Defensive top-level `event` guard — NULL or non-object input coerced to `{}` before any work
- **#6** Explanatory down-migration comment for why USAGE on schema `profile` is NOT revoked (v1 needs it)
- **#7** `USING ERRCODE = '22023'` (invalid_parameter_value) on all RPC validation RAISEs — stable SQLSTATE for axum-kbve callers
- **#8** Tighten `discord_provider_id` regex `{15,25}` → `{17,20}` (matches Discord snowflake; brand new table so no legacy concern)
- **#9** Table COMMENT updated to reflect SELECT-only + rationale

**Deferred**
- **#5** RPC canonicalization smoke in migration — needs `auth.users` write; lives in CI pgTAP suite (consistent with all prior rounds)

**Local verification — 7 real-data probes all green**

| Probe | Check | Result |
|---|---|---|
| A_tighter_regex | 15-digit provider_id | SQLSTATE 22023, "invalid discord_provider_id" |
| B_too_long | 21-digit provider_id | SQLSTATE 22023 |
| C_null_user | NULL `user_id` | SQLSTATE 22023 ✓ |
| D_future_ts | `+5min` `refreshed_at` | SQLSTATE 22023 ✓ |
| E_null_event | `hook(NULL)` | `{"claims":{}}` (defensive guard fires) |
| F_scalar_event | `hook('"scalar"')` | `{"claims":{}}` (defensive guard fires) |
| G_strip_inbound | valid user + smuggled `owned_guilds=["attacker_smuggled"]` | output is canonical cached snowflakes only (strip+set works) |

Rollback drops table + RPC + helper cleanly; preserves USAGE on schema `profile` for v1 hook (verified post-rollback).

**7 review rounds done, ~94 disposition points addressed.** The cross-swap race fix in #4 is the only real correctness change this round — everything else is hardening polish. Hook + RPC + helper are now all defended at every boundary I can think of.

**Follow-ups still owned outside this PR**
- pgTAP test suite for canonicalisation + LIMIT-1-on-unnest regression (CI-only)
- `dev-docker-compose.yml` → `ghcr.io/kbve/postgres:17.4.1.069-kilobase` for prod-parity local dev

h0lybyte:
Round-8 fold-in pushed (commit 10f781f8c). All 6 points taken; no declines.

Reviewer explicitly noted "this migration looks ready" — these are final polish items.

**Taken (6/6)**
- **#1** Event guard added to the v1 down hook too. The down hook comment claimed v2 hardening preserved but was missing the round-7 guard — fixed.
- **#2** Toned down the throughput comment on the coarse advisory lock. Replaced "well within Postgres advisory-lock contention budgets" with a more honest framing: acceptable at OAuth/refresh rates, but if write volume grows the alternative is deterministic row-level locking.
- **#3** Inline migration note documenting the row-lock alternative (with the caveat that row locks alone don't cover the "row does not yet exist" case the advisory lock handles).
- **#4** Fixed misleading `service_role` SELECT justification. The previous comment claimed SELECT was needed for "ON CONFLICT visibility" — but the RPC runs `SECURITY DEFINER` as `postgres`, so ON CONFLICT is satisfied by the RPC owner's privileges. Real reason is dashboard freshness peeks (P5.8c). Corrected the inline `GRANT` comment, the table `COMMENT`, and the privilege-assertion message.
- **#5** Added scalar-event + NULL-event smoke checks (smoke 5, 6) to the hook smoke block. Regression guard for the round-7 top-level event guard.
- **#6** Hook owner assertion: `public.custom_access_token_hook` must be owned by `supabase_auth_admin`. Pairs with the existing RPC owner assertion (round 6 #9). Important because the INVOKER helper relies on running under the hook's SECURITY DEFINER owner — drift would silently emit empty claims.

**Local verification — apply + rollback + apply cycle**

Apply: clean (21ms); two new smoke checks + hook owner assertion all passed during apply.

Post-rollback verification of #1 (v1 hook now carries the guard):

| Probe | Check | Result |
|---|---|---|
| v1_null_event | `jsonb_typeof(hook(NULL))` | `'object'` ✓ |
| v1_scalar_event | `jsonb_typeof(hook('"scalar"'))` + claims type | both `'object'` ✓ |
| v1_object_normal | inbound `owned_guilds=['123']` claim | stripped; output `{"role":"authenticated"}` (v1 hardening preserved) |

Re-apply clean.

**8 review rounds done, ~100 disposition points addressed.** The reviewer signed off after this round — every concern across all 8 passes is either folded in, deferred to a tracked CI follow-up, or declined with reasoning.

**Follow-ups still owned outside this PR**
- pgTAP test suite for canonicalisation + LIMIT-1-on-unnest regression (replaces the per-round in-migration smokes that keep getting deferred)
- `dev-docker-compose.yml` → `ghcr.io/kbve/postgres:17.4.1.069-kilobase` for prod-parity local dev

h0lybyte:
Round-9 fold-in pushed (commit f10ff56ec). All 6 points taken; no declines.

**Taken (6/6)**
- **#1** Two stale comments still described the round-6 per-key locks on `(user_id, provider_id)` — both rewritten to reflect the round-7 single coarse namespace lock.
- **#2** Added `GRANT USAGE ON SCHEMA profile TO service_role` + matching assertion. Function + table grants alone fail with "permission denied for schema profile" without it; idempotent grant + assertion to catch future drift.
- **#3** Defense-in-depth negative assertions: anon and authenticated must NOT have direct INSERT/UPDATE/DELETE on the cache. REVOKE ALL covers it; assertions catch future drift.
- **#4** Explicit comment above `ENABLE ROW LEVEL SECURITY` documenting why FORCE RLS is intentionally NOT used — postgres-owned SECURITY DEFINER RPC bypasses via owner anyway; revisit if RPC owner moves or non-superuser caller gains DML.
- **#5** Extended NULL-event smoke (smoke 6) with claims-shape assertion, matching the scalar-event smoke (smoke 5).
- **#6** Explicit `p_user_id IS NOT NULL` guard in the SQL helper. Logically unnecessary; documents intent + protects against planner edge cases on future index/partition changes.

**Local verification — apply + rollback + apply cycle**

Apply: clean (18ms); all new assertions + smoke checks passed during apply.

Manual probes of new behaviour:

| Probe | Check | Result |
|---|---|---|
| service_role_profile_usage | `has_schema_privilege('service_role','profile','USAGE')` | `t` ✓ (round 9 #2 grant landed) |
| anon_insert | `has_table_privilege('anon','...','INSERT')` | `f` ✓ (REVOKE ALL survives) |
| helper(NULL, NOW()) | `[]` | NULL guard returns empty cleanly |

Rollback drops table + RPC + helper cleanly; preserves schema USAGE for v1 hook. Re-apply clean.

**9 review rounds done, ~106 disposition points addressed.** Reviewer signed off in round 8 ("looks ready") and round 9 ("very solid shape") — these last two rounds are pure polish.

**Follow-ups still owned outside this PR**
- pgTAP test suite for canonicalisation + LIMIT-1-on-unnest regression (replaces deferred in-migration smokes)
- `dev-docker-compose.yml` → `ghcr.io/kbve/postgres:17.4.1.069-kilobase` for prod-parity local dev

h0lybyte:
Round-10 fold-in pushed (commit a0612c04d). 6 of 11 taken, 1 declined, 2 deferred, 2 no-ops.

**Standout — privilege tightening**
- **service_role no longer has direct SELECT on the cache table** (#1). Earlier rounds granted it for dashboard freshness peeks, but that exposed every cached row's `discord_provider_id` + `owned_guilds` list to broad service-role reads. Replaced with a narrow SECURITY DEFINER RPC `profile.service_get_discord_bootstrap_cache_freshness(uuid)` returning only `refreshed_at + updated_at + created_at + guild_count` for the requested user. Dashboard (P5.8c) reads through the RPC; the underlying guild list never leaves the SECURITY DEFINER boundary.

**Standout — observability**
- `service_upsert` RPC now `RAISE LOG`s on two events that were previously silent:
  - **Account relinks** → `"relinked discord_provider_id=X from user=A to user=B"`
  - **Input filtering** → `"filtered N of M owned_guilds (NULL/non-snowflake/dup/cap) for user=X"`
- `RAISE LOG` honours `log_min_messages` so the messages stay out of client output in normal operation but show up in `pg_log` for ops review.

**Taken (6/11)**
- **#1** Narrow freshness RPC + revoke direct service_role SELECT
- **#3** Capture `v_relinked_user_id` via RETURNING + `RAISE LOG` (partial — no return-type change, see #10 decline)
- **#4** Keep silent canonicalisation filter (one bad guild from Discord's API shouldn't break OAuth refresh) but `RAISE LOG` when filtering
- **#5** Inline comment near ON CONFLICT documenting that same-user `discord_provider_id` replacement is supported
- **#6** `jsonb_agg(...)` over `to_jsonb(array_agg(...))` — one aggregation pass instead of two
- **#7** New smoke 7: valid UUID + no cache row + attacker-supplied `owned_guilds` claim → output must be `[]`

**Declined**
- **#10** Return TABLE instead of void — premature signature change; `RAISE LOG` from #3 + #4 gives observability without breaking the API

**Deferred**
- **#2** `refreshed_at` index — same as every prior round; no janitor/global-scan planned
- **#8** RPC canonicalisation smoke inside migration — needs `auth.users` write; lives in CI pgTAP

**No-op**
- **#9** Caution about `transaction:false` + smoke sections — commentary only
- **#11** Normalise to child table — reviewer explicitly "do not normalize prematurely"

**Local verification — 5 real-data probes all green**

| Probe | Check | Result |
|---|---|---|
| A_freshness_rpc | RPC returns metadata only | `refreshed/updated/created + guild_count=2` (no provider_id or guild list leaked) |
| B_freshness_missing | RPC for unknown user | `0 rows` ✓ |
| C_service_role_no_select | `has_table_privilege('service_role', ..., 'SELECT')` | `f` ✓ (round-10 revoke confirmed) |
| D_relink_completed | relink succeeds | `LOG: relinked discord_provider_id=... from user=A to user=B` visible |
| E_filter_completed | 4 input → 2 stored | `LOG: filtered 2 of 4 owned_guilds (NULL/non-snowflake/dup/cap) for user=...` visible |

Rollback drops freshness RPC + helper + upsert RPC + table + validator cleanly. Re-apply green.

**10 review rounds done, ~117 disposition points addressed.** PR is locked in. The freshness-RPC change tightens the service-role surface materially — direct table SELECT is now zero, and writes have always been zero. All cache access flows through one of two narrow RPCs.

**Follow-ups still owned outside this PR**
- pgTAP test suite for canonicalisation + LIMIT-1-on-unnest regression (CI-only, deferred consistently across all 10 rounds)
- `dev-docker-compose.yml` → `ghcr.io/kbve/postgres:17.4.1.069-kilobase` for prod-parity local dev

h0lybyte:
Round-11 fold-in pushed (commit 54560e67c). 7 of 10 taken (1 required, 6 polish), 3 no-ops.

**Required fix landed**
- **#1** `service_role_select` RLS policy dropped entirely. After round-10 revoked direct table SELECT, the policy was inert but a drift risk: a future re-grant would immediately reactivate full-table reads through the stale policy. Now both the policy and the grant are gone; **#10** drift assertion blocks reintroduction (`RAISES` if any `service_role` policy reappears on the cache table).

**Standout — log hygiene**
- `RAISE LOG` messages trimmed to avoid leaking sensitive identifiers into `pg_log` (which feeds ClickHouse via Vector):
  - Raw `discord_provider_id` removed from the relink log
  - `user_id` removed from the filter log (count alone is the actionable signal)
  - Relink log keeps before/after `user_id`s — that pair is the actionable info for incident response

**Standout — caller ergonomics**
- Freshness RPC switched to LEFT JOIN over a single-row input — always returns exactly one row (NULL timestamps + `guild_count=0` when no cache exists). Saves the Axum/dashboard side an empty-result-set special case.

**Other taken**
- **#2** Updated stale "dashboard SELECT peek" comments — dashboard reads via the freshness RPC now
- **#4** Split filter log into two distinct signals: `"filtered N invalid"` (Discord API sent bad data) vs `"deduped/capped M"` (our canonicalisation collapsed dupes or hit cap). Makes upstream weirdness easier to spot.
- **#6** `ROWS 1` hint added to the freshness RPC. The helper is scalar-returning so `ROWS` isn't valid there — caught by local apply (`"ROWS is not applicable when function does not return a set"`) and removed.

**No-ops (3/10)**
- #7 validator + cap CHECK redundancy — reviewer says keep both
- #8 canonicalisation subquery — reviewer says leave
- #9 down policy cleanup — DROP TABLE cascades, already correct

**Local verification — 5 real-data probes all green**

| Probe | Check | Result |
|---|---|---|
| A_missing_user_row | freshness RPC for unknown user | 1 row, NULL timestamps, `guild_count=0` ✓ |
| B_no_service_role_policy | `count(*)` of service_role policies | `0` ✓ (round-11 drop confirmed) |
| C_split_logs | 6 inputs (2 invalid, 1 dup, 3 valid) | Two distinct LOG lines: `"filtered 2 invalid (of 6 input)"` + `"deduped/capped 1 (4 valid -> 3 stored)"` |
| D_clean_relink_log | log message | `"relinked discord provider from user=A to user=B"` — no provider_id leaked |
| E_drift_assertion | simulated policy reintroduction | drift logic correctly detects it |

Rollback drops freshness RPC + helper + upsert RPC + table + validator cleanly. Re-apply green.

**11 review rounds done, ~127 disposition points addressed.** The service_role surface is now strictly zero direct table privileges + zero RLS policy — all reads go through one of two narrow SECURITY DEFINER RPCs (`service_upsert` for writes, `service_get_..._freshness` for dashboard metadata). The drift assertion locks both invariants in place against future PRs.

**Follow-ups still owned outside this PR**
- pgTAP test suite for canonicalisation + LIMIT-1-on-unnest regression (CI-only)
- `dev-docker-compose.yml` → `ghcr.io/kbve/postgres:17.4.1.069-kilobase` for prod-parity local dev

h0lybyte:
Round-12 fold-in pushed (commit 26cc8337d). 3 of 6 taken, 2 declined w/ reasoning, 1 no-op.

**Taken**
- **#1** Drift guard rewritten to use `'service_role' = ANY (roles)` instead of `roles @> ARRAY[...]::name[]`. Cleaner.
- **#2** New positive assertion: `supabase_auth_admin_select` policy must exist on the cache (FOR SELECT, role `supabase_auth_admin`). Pairs with the round-11 #10 negative drift guard — together they pin the RLS surface to exactly one policy + role + cmd combination.
- **#5** Tightened freshness RPC comment to be precise about NULL: non-NULL input always yields one row; NULL input returns zero rows.

**Declined w/ reasoning**
- **#3** Drop `user_id`s from relink RAISE LOG. Reviewer explicitly says "I would not block merge on this." Relink events are rare (account moves); when they happen, knowing which users were involved is the actionable signal for incident response. `provider_id` (higher-sensitivity) was already trimmed in round 11. Keep user_ids.
- **#4** `ROWS 1` on `profile.get_discord_owned_guilds_claim`. Already attempted in round 11; Postgres rejected with `"ROWS is not applicable when function does not return a set"` — the helper `RETURNS jsonb` (scalar), not `SETOF`/`TABLE`. `ROWS` is only valid for set-returning functions. The freshness RPC (which DOES return TABLE) has `ROWS 1`.

**No-op**
- **#6** `refreshed_at` index — reviewer explicitly says "skip it"

**Local verification — apply + rollback + apply cycle**

Apply clean (18ms); both new assertions passed during apply (no service_role policy + supabase_auth_admin_select exists).

`pg_policies` snapshot post-apply:

```
policyname                 | cmd    | roles
---------------------------+--------+-----------------------
supabase_auth_admin_select | SELECT | {supabase_auth_admin}
```

Exactly one policy, one role, SELECT only — the pinned state. Rollback + re-apply clean.

**12 review rounds done, ~133 disposition points addressed.** The RLS surface is now locked: exactly one policy must exist (positive assertion), and no `service_role` policy may exist (negative assertion). Both fire at migration time.

**Follow-ups still owned outside this PR**
- pgTAP test suite for canonicalisation + LIMIT-1-on-unnest regression (CI-only, deferred consistently)
- `dev-docker-compose.yml` → `ghcr.io/kbve/postgres:17.4.1.069-kilobase` for prod-parity local dev

PR #11572, opened 2026-06-01
feat(jedi+axum-kbve): PgCluster (CNPG-aware) + GET /api/v1/wallet/me/ledger

h0lybyte:
## Summary

Establish the Rust-canonical Postgres core (`jedi::PgCluster`) and prove it out with the first read endpoint moved off the SQL `proxy_*_readonly` family: `GET /api/v1/wallet/me/ledger`.

The original SQL-first ledger proxy committed earlier on this branch has been removed in the final commit; the Rust handler is the new source of truth for caller-scoped market history reads, with future read RPCs to follow the same pattern.

## Commits in this PR

1. **`feat(jedi): PgCluster — CNPG-aware Postgres pool set, feature-gated`** (`57ea29e`) — new `jedi::state::pg` module, three-pool wrapper (`rw` / `ro` / `any`) matching the `supabase-cluster` CNPG topology, `tokio-postgres-rustls` connector (transparent under `sslmode=disable`), env-driven config + per-pool knobs, `KBVE_PG_USE_POOLER` toggle, `with_caller_read(sub, role, closure)` JWT impersonation primitive. Gated behind `jedi.features.postgres`. 6 unit tests pass.

2. **`feat(axum-kbve): GET /api/v1/wallet/me/ledger via jedi PgCluster; drop SQL proxy`** (`26f6399`) — wires `PgCluster::from_env()` into axum-kbve startup, registers `/me/ledger`, removes the SQL migration + test + schema-mirror entry that previously lived on this branch.

## Three-pool topology

| Role | CNPG service                    | Use |
|------|---------------------------------|-----|
| `rw`  | `supabase-cluster-rw` | primary; writes + read-after-write |
| `ro`  | `supabase-cluster-ro` | HA replicas; strict reads |
| `any` | `supabase-cluster-r`  | any-healthy fallback for read paths |

`any` exists so reads survive a replica blip without falling through to the primary, and gives forward-compat with 5-instance clusters.

## Handler shape

```
GET /api/v1/wallet/me/ledger
    ?limit=50
    &before_created_at=2026-06-01T17:38:47Z
    &before_id=12345
    &source_kinds=market_buy,market_sell,market_fee   # default; pass `all` to widen
    &currency=khash                                    # optional; omit for both
Authorization: Bearer <user JWT>
```

Response:
```json
{
  "rows": [ { "ledger_id": …, "currency": "khash", "delta": …, "balance_after": …,
              "source_kind": "market_buy", "reason": …, "ref_type": …, "ref_id": …,
              "created_at": "2026-06-01T17:38:47+00:00" }, … ],
  "next_cursor": { "before_created_at": "…", "before_id": … }    // null on last page
}
```

Validation:
- Half cursor (one of `before_created_at` / `before_id` supplied without the other) → `400 invalid_cursor`.
- `before_id <= 0` → `400 invalid_cursor`.
- `source_kinds=""` → `400 invalid_source_kinds` (omit the param for defaults; pass `all` to widen).
- `currency` ∉ {credits, khash} → `400 invalid_currency`.

Auth:
- `resolve_user(headers)` extracts the JWT sub (matches existing `/me/balance`, `/me/coupons`, `/me/redeem-coupon` pattern).
- `PgCluster::with_caller_read(sub)` opens a tx on the `ro` pool, sets `request.jwt.claims` via `set_config(LOCAL)`, runs the query under the `WHERE account_id = (SELECT id FROM wallet.account WHERE user_id = auth.uid() AND kind='user')` constraint, commits. Server cannot impersonate another user via a query-string trick — the row scope is enforced inside the SQL.

## Why drop the SQL proxy?

Rust-canonical = one source of truth per read path, no SQL drift, types live with the handler, easier to test, easier to evolve. PostgREST-direct fallback (Layer 1 in the original three-transport plan) is being retired in favor of the Rust path; if we want a fallback later we can add it under a clean abstraction rather than maintaining two query implementations.

## Build / verification

- [x] `cargo build -p jedi --no-default-features` — no-feature build green (vestigial tokio-postgres dep is now opt-in)
- [x] `cargo build -p jedi --no-default-features --features postgres`
- [x] `cargo test -p jedi --no-default-features --features postgres --lib state::pg::tests` — 6/6 pass (env parsing, compat-key fallbacks, pooler host rewrite per role, `KBVE_PG_USE_POOLER` truthy/falsy)
- [x] `cargo check -p axum-kbve` — green with `jedi.features += ["postgres"]`
- [ ] axum-kbve smoke against `dev-docker-compose.yml` (fire the route, seed a couple `wallet.ledger` rows via `wallet.service_credit`, verify the response shape + cursor advance + currency filter) — pending local run
- [ ] CNPG cluster smoke (rw / ro / any pool acquire) against a real `supabase-cluster` namespace — pending k8s round

## Follow-ups (out of scope here)

- Port existing `WalletClient` read endpoints (`/me/balance`, `/me/coupons`) off `diesel-async` onto `PgCluster::with_caller_read`. Then drop diesel-async from `kbve.wallet`.
- Sweep remaining `kbve.*` modules (guild, referral, characters, …) off diesel-async incrementally.
- Decide whether the existing PostgREST `proxy_market_my_{listings,bids}_readonly` proxies stay or get retired the same way.


h0lybyte:
GPT review applied. Triage:

**Fixed in 3e1a006**
- **#2 empty array** → reject `'{}'` with `22023`. Empty array is almost always a client bug; explicit error directs callers to pass `NULL` if they want full-ledger widening.
- **#7 non-positive `p_before_id`** → reject with `22023`.
- **#6 validation order** → moved `private.proxy_market_caller_account()` resolution after the argument checks so cursor/array errors surface deterministically instead of getting masked by `WLT01`.
- **#1 ref leak (header contradiction)** → kept `ref_type`/`ref_id` in the return shape, dropped the misleading header line. The `ref_id` on `market_*` rows points to the caller's own bid/listing, which is already publicly browseable via `proxy_market_list_active_readonly` / `proxy_market_listing_detail_readonly`. Header now spells out the rationale plus documents `reason` as enum-like service-emitted strings (`'buy_now'`, `'listing_cancelled'`, `'expired_with_bid'`, …) — no admin notes or freeform internal payloads.

**Decided: keep current shape**
- **#5 split into `proxy_wallet_my_ledger_readonly`** — passing on it. Caller-scoped, no cross-user leak, no separate auth boundary needed. The default array gives "market history" out of the box; `NULL` widens to the caller's full ledger. Documented explicitly. If we later want a hard auth boundary (e.g. restricting non-market reads to a different role), we can carve out a sibling proxy then.
- **#8 reason normalization** — checked the call sites. `reason` is written by `wallet.service_credit` / `service_debit` / `service_settle_listing` / `service_cancel_listing` and the marketplace RPCs, with stable enum-like values. No admin notes leak in. Safe to expose as-is.
- **#9 `STABLE` vs `VOLATILE`** — keeping `STABLE`. Matches the existing `my_listings` / `my_bids` proxies. `private.proxy_market_caller_account()` is also `STABLE` and only reads `auth.uid()` + a wallet account lookup, no provisioning.
- **#10 `has_more` flag** — UI derives from `LIMIT v_limit + 1`-style client pagination off the last row's `(created_at, ledger_id)`. Adding `has_more` per row to a `RETURNS TABLE` shape would be ugly; skipping.

**Deferred to a follow-up PR (with EXPLAIN evidence)**
- **#3 partial covering indexes** + **#4 `ANY(array)` not matching the partial-index predicate**. The marketplace-only hot path probably wants:
  ```sql
  CREATE INDEX CONCURRENTLY idx_wallet_ledger_account_market_cursor
      ON wallet.ledger (account_id, created_at DESC, id DESC)
      INCLUDE (currency, delta, balance_after, source_kind, reason, ref_type, ref_id)
      WHERE source_kind IN ('market_buy', 'market_sell', 'market_fee');
  ```
  Want to land this against the existing `wallet_ledger_account_currency_idx` first and measure with real ledger volume before deciding between (a) partial covering index + leaving the proxy as-is, (b) branching to a hardcoded `IN (...)` query so the planner can prove the partial-index predicate match, or (c) both. Will open a separate PR once we have measurements.

Header in the migration now documents the deferred index work so reviewers of the follow-up PR have the context.

h0lybyte:
Round 2 triage. Pushed in `d5b88ce`.

**Applied to migration**
- **#1 `CREATE OR REPLACE` return-shape safety** → migration now prepends `DROP FUNCTION IF EXISTS ... CASCADE`-style guard (no `CASCADE` — explicit drop with the exact signature) before the `CREATE OR REPLACE` so future signature edits to the return shape don't trip the "cannot change return type" check. No-op on first apply.
- **#4 `ROWS 50` → `ROWS 100`** → reflects the actual contract max after the limit clamp.

**Applied as a regression test file (#11)**
Added [`packages/data/sql/dbmate/migration-tests/20260601173847_wallet_marketplace_my_ledger_proxy.test.sql`](packages/data/sql/dbmate/migration-tests/20260601173847_wallet_marketplace_my_ledger_proxy.test.sql) mirroring `20260516015242_wallet_marketplace_proxies.test.sql`. Fixture seeds Alice with reward + coupon + market_buy x3 (distinct ts) + market_sell + market_fee + credits-currency market_buy + three market_buy rows sharing a `created_at` (written via direct `INSERT INTO wallet.ledger` using `wallet.replay_fingerprint(...)` so the trigger is satisfied) and Bob with one private market_buy row.

Cases covered, in test order:
1. Anon → `28000`
2. Authenticated orphan (no `wallet.account`) → `WLT01`
3. Default `p_source_kinds` excludes non-market rows + every returned row is Alice's
4. `p_source_kinds := NULL` includes reward + coupon rows, still no Bob leak
5. Empty array `'{}'` → `22023`
6. Half cursor (only one of the pair) → `22023`
7. Zero / negative `p_before_id` → `22023`
8. Equal `created_at`: `(created_at, id)` cursor strictly advances through the three tied rows — no dup, no gap (this is the case GPT flagged as "especially important")
9. `p_currency` filter applied + caller-scoped
10. Custom `p_source_kinds` narrows to `market_sell` only without widening account scope
11. Limit clamp (sibling-proxy policy, not rejection)

**Audited inline (notes only)**
- **#7 helper hardening** → `private.proxy_market_caller_account()` (in migration `20260516015242_wallet_marketplace_proxies`) is `SECURITY DEFINER`, `STABLE`, `SET search_path = ''`, fully qualified refs, no dynamic SQL, raises `28000` on `auth.uid() IS NULL` and `WLT01` on missing wallet account. Lives in `private` schema (PostgREST does not expose). `EXECUTE` granted to `service_role` only; `PUBLIC` + `anon` + `authenticated` `REVOKE`d. The wrapper's empty `search_path` is therefore additive, not load-bearing on its own.
- **#8 reference ownership invariant** → confirmed across `20260515054304_wallet_marketplace_rpcs` + `20260520124536_wallet_listing_settle_inventory`:
  - `market_buy` → debit/credit on the bidder/buyer account (i.e. the caller), `ref_type='bid'`, `ref_id` = caller's own bid. (Refund credits on prior-bid outbid go to that prior bidder's account, with the same ref to *their* own bid.)
  - `market_sell` → credit on the seller account (i.e. the caller), `ref_type='listing'`, `ref_id` = caller's own listing — also publicly browseable via `proxy_market_list_active_readonly` / `proxy_market_listing_detail_readonly`.
  - `market_fee` → credit on `wallet.treasury_account_id()`, `ref_type='listing'`, `ref_id` = listing being settled. Treasury has no `user_id`/`auth.uid()` binding, so `market_fee` rows never appear in a regular caller's response — the default filter includes it for forward compat with a future treasury-admin caller, not for leakage. Documented inline in the migration header.
- **#6 `service_role` grant rationale** → matches every existing read-only marketplace proxy. Direct service-role invocations hit `private.proxy_market_caller_account()` and raise `28000` because `auth.uid()` is `NULL`. Grant stays for sibling-symmetry + ops/migration test access; the function does **not** accept a target user id. Spelled out in the header for #6's "explicitly document the dependency on caller context" ask.
- **#10 RLS** → `wallet.ledger` has no RLS policies (gated entirely by definer proxies); the proxy is `SECURITY DEFINER OWNER service_role`, so reads bypass RLS by design. The caller cannot influence `v_account` — `auth.uid()` comes from the JWT, not from the function arguments.

**Decided keep (no change)**
- **#5 `p_limit` validation** → kept the `LEAST(GREATEST(COALESCE(p_limit, 50), 1), 100)` clamp. Sibling proxies (`my_listings_readonly`, `my_bids_readonly`) use the identical pattern; matching their policy is more valuable than tightening just this one proxy. Test #11 documents the behavior.
- **#9 `reason` as enum/domain** → real concern, but converting `wallet.ledger.reason` from `TEXT` to an enum/domain touches every existing ledger row and every service-credit / service-debit call site. Out of scope for this PR. Filed mentally for a follow-up; for now the migration header documents `reason` as a stable enum-like vocabulary and the test fixtures pin the expected values for the existing call sites.

**Deferred to the follow-up index PR**
- **#2 generic-plan risk** + **#3 partial-index predicate match for `ANY(p_source_kinds)`**. Will land:
  1. A `CREATE INDEX CONCURRENTLY` covering partial index on the marketplace trio.
  2. `EXPLAIN (ANALYZE, BUFFERS)` evidence for the 5 query shapes GPT listed (default+no-currency, default+single-currency, custom single source kind, `NULL` source kinds, deep cursor).
  3. If the planner refuses to match `ANY(p_source_kinds)` against the partial predicate, branch the default path inside the proxy to a literal `IN (...)` query (custom-array + `NULL` paths stay on the generic predicate).
  Header in this PR's migration already documents that follow-up so reviewers there have context.

h0lybyte:
Round 3 pushed in `af9087f`. Mostly reverting overreach from round 2.

**Applied**
- **#1 + #2 revert the upfront `DROP FUNCTION`** → restored plain `CREATE OR REPLACE`. The drop was misguided defensive plumbing: it discards grants/comments/dependent-object linkage and defeats the point of `CREATE OR REPLACE`. A future migration that actually changes the return shape will manage its own drop deliberately at that time; this migration can't preemptively help it. The earlier rationale comment was wrong — removed.
- **#4 softened the index-backing claim** → header now spells out that `(account_id, currency, created_at DESC, id DESC)` aligns for a specified currency but can't produce a globally sorted result when `p_currency` is `NULL` without extra work. Index-PR follow-up still flagged.
- **#7 `ROWS 100` → `ROWS 50`** → agreed, `ROWS` is an estimated avg per invocation, not a declared upper bound. Default page size + UI default both point at 50.
- **#9 attribute formatting** → split `LANGUAGE plpgsql STABLE SECURITY DEFINER SET search_path = ''` onto separate lines so the security posture is grep-friendly for audits.
- **#10 ACL assertion** → added test case `#0` calling `has_function_privilege('authenticated' / 'anon', signature, 'EXECUTE')` and asserting `(true, false)`. Belt-and-suspenders against accidental privilege regressions on a re-run.

**Audited inline (no code change)**
- **#8 ref ownership across all `market_*` write sites** — traced every call site in `20260515054304_wallet_marketplace_rpcs` and `20260520124536_wallet_listing_settle_inventory`:
  - `market_buy` / escrow at `service_place_bid`: account = bidder (caller), `ref_type='listing'`, `ref_id=p_listing_id`. Public listing.
  - `market_buy` / escrow at `service_buy_now`: account = buyer (caller), `ref_type='listing'`, `ref_id=p_listing_id`. Public listing.
  - `market_buy` / refund at `wallet.refund_active_bid`: account = prior bidder (caller of the credit context), `ref_type='bid'`, `ref_id=v_bid_id` — the bid being refunded is owned by the credit recipient. Caller's own.
  - `market_sell` at `wallet.distribute_settlement`: account = seller (caller), `ref_type='listing'`, `ref_id=p_listing_id`. Seller's own listing, also publicly browseable.
  - `market_fee` at `wallet.distribute_settlement`: account = `wallet.treasury_account_id()`, `ref_type='listing'`, `ref_id=p_listing_id`. Treasury has no `user_id` binding so this row is unreachable via the caller-scoped proxy in normal use.

  GPT's hypothetical "seller-side credit referencing the buyer's winning bid" doesn't materialize — the actual settlement code references the **listing**, not the winning bid. Every `ref_id` returned by this proxy is either the caller's own object or a publicly browseable listing. Invariant holds; documented in the migration header.

**Decided keep**
- **#5 limit clamp vs strict validation** — sibling proxies (`my_listings_readonly`, `my_bids_readonly`) clamp identically. Consistency wins; tightening just this one proxy creates a contract drift.
- **#3 + #6 future branching** → still deferred to the index PR. The literal-`IN (...)` branch only earns its keep once the partial index exists and `EXPLAIN` confirms the planner refuses to match `ANY(p_source_kinds)`. Doing the branch now would be speculative.

PR #11753, opened 2026-06-04
docs(sql/schema): mirror profile.discord_bootstrap_cache + custom_access_token_hook (post-#11488)

h0lybyte:
## Summary
- Backfills the per-schema review surface under \`packages/data/sql/schema/profile/\` to match the canonical state established by [#11488](https://github.com/KBVE/kbve/pull/11488)
- Two new idempotent-safe files; no behavioural change, no migration

## Why
PR #11488 merged the dbmate migration but the hand-authored \`schema/\` mirror was left out of date. That tree is the per-schema review surface per [schema/README.md](packages/data/sql/schema/README.md) and is supposed to reflect what the live DB looks like.

## New files
| File | Contents |
|---|---|
| \`schema/profile/profile_discord_bootstrap_cache.sql\` | validator fn, cache table + constraints + RLS + the single \`supabase_auth_admin_select\` policy, \`service_upsert\` RPC, \`service_get_..._freshness\` RPC, \`get_discord_owned_guilds_claim\` helper |
| \`schema/profile/profile_custom_access_token_hook.sql\` | \`public.custom_access_token_hook\` v2 (top-level event guard, claims-object typecheck, invalid-user_id privileged-claim stripping, explicit \`owned_guilds\` strip before set, delegation to the helper) |

Both follow the schema/ convention:
- Idempotent-safe (`CREATE TABLE IF NOT EXISTS`, `CREATE OR REPLACE FUNCTION`, `DROP POLICY IF EXISTS`)
- Tables → indexes → policies → functions → grants
- File prefix matches the schema name for glob-load alphabetical order

## Mirror fidelity vs PR #11488 migration
- Cache table column definitions, constraints (FK / UNIQUE / size_cap / validator / timestamps_ordering_and_floor), and comment match exactly
- Validator function signature + body + revokes match (no caller EXECUTE; CHECK constraint runs it as table owner)
- \`service_upsert\` RPC body matches: coarse advisory lock on the relink namespace, split filter logs (without user_id), relink log (without provider_id), SQLSTATE 22023 on validation failures, explicit NULL \`p_refreshed_at\` handling, owner pinned to postgres
- \`service_get_..._freshness\` RPC matches the round-11 LEFT JOIN shape (always one row for non-NULL \`p_user_id\`; zero rows for NULL) + \`ROWS 1\` planner hint
- \`get_discord_owned_guilds_claim\` helper matches the round-10 \`jsonb_agg\` form + round-9 \`p_user_id IS NOT NULL\` guard + defensive NULL/regex filter
- \`public.custom_access_token_hook\` v2 body matches the post-round-9 hardened form

## Not changed
- \`schema/profile/username.sql\` — PR #11488 did not touch \`profile.username\` (the supabase_auth_admin SELECT grant came from migration \`20260510210000\`, independent of this PR)
- Other schema directories — out of scope

## Test plan
- [x] Files lint clean — no syntax errors in the SQL surface (review-only; not applied)
- [x] Glob-load ordering: \`profile_custom_access_token_hook.sql\` < \`profile_discord_bootstrap_cache.sql\` < \`username.sql\`. PL/pgSQL late-binding means the hook's reference to \`profile.get_discord_owned_guilds_claim\` resolves at call time, not create time
- [ ] Reviewer spot-check: every constraint, GRANT, REVOKE, policy in the merged migration is reflected in one of the two files


h0lybyte:
Expanded PR with nx smoke targets so future migration work can verify the full chain with one command.

**New files**
- `packages/data/sql/dbmate/smoke.sh` — lifecycle script (volume nuke → compose up → wait healthy → `dbmate up` → status → optional `--rollback` → optional `--keep`). Knows both stacks; defensively tears the other one down first so port 54322 is always free.

**Updated**
- `packages/data/sql/project.json` — 6 new targets:
  - `smoke-vanilla` / `smoke-kilobase` / `smoke` (both, sequential)
  - `smoke-vanilla-keep` / `smoke-kilobase-keep` (leave stack up for psql probing)
  - `kilobase-up` / `kilobase-down` / `kilobase-reset` (lifecycle wrappers paralleling existing `dev-*`)
- `packages/data/sql/README.md` — quick-start now points at the nx targets
- `packages/data/sql/dbmate/README.md` — Local Testing section reorganised; smoke targets table first, manual sequence below as reference

**Local verification**
- `nx show project data-sql` lists all 14 targets after `nx reset`
- `nx run data-sql:smoke-vanilla` → 57/57 migrations apply against fresh `postgres:17-alpine`, status clean, teardown OK ✓
- `nx run data-sql:smoke-kilobase` → 57/57 migrations apply against `ghcr.io/kbve/postgres:17.4.1.069-kilobase`, real `auth.*` schema present (not stub), teardown OK ✓
- Both targets defensively tear down the other stack first; back-to-back switching works without manual cleanup.

**Use case for PR #11488 itself**: had this been in place earlier, the 12 review rounds could have re-verified the full chain on both stacks via `nx run data-sql:smoke` in seconds instead of running ad-hoc `dbmate up/rollback/up` cycles.

**Follow-ups (out of scope here)**
- Wire `smoke` into `ci-dbmate-deploy.yml` as a pre-deploy gate
- Generalize the `smoke.sh` pattern if other DB-adjacent stacks (n8n, forgejo) need it

PR #12364, opened 2026-06-14
fix(cryptothrone,irc-gateway): route realm chat to #general

h0lybyte:
## Bug
In-game chat didn't reach IRC. Root cause: cryptothrone realm chat was routed to **`#cryptothrone`** — an empty channel — instead of **`#general`** where IRC users actually are. The earlier local-echo fix (#12345) made the sender see their own message, which **masked** that it never reached the real channel.

Not a serialization bug — the frame parses fine (`kind:"chat"` → `MessageKind::Chat` via snake_case; `payload` is `#[serde(default)]`). Purely the channel.

## Fix — `#cryptothrone` → `#general`, in sync across all four sites
- gateway `GAME_PROFILES` — gamechat session joins `#general`
- gateway `HISTORY_CHANNELS` — ring-buffer listener tracks `#general`
- client `REALM_CHAT_CHANNEL` (net-config) — sends + filters on `#general`
- zone registry `channel` (zones.ts) — kept consistent for future WorldScene routing

All four must agree (gateway whitelist-drops mismatched channels), hence the coordinated change.

## Versions
- cryptothrone `0.1.22 → 0.1.23`, irc-gateway `0.1.22 → 0.1.23`

## Verified
- gateway `cargo build` clean; history tests unaffected (sanitize helper tests are channel-agnostic).
- Needs a live check after deploy: send in-game → appears in `#general` for IRC clients + new joiners get backscroll.

### DIFF
--- a/apps/cryptothrone/astro-cryptothrone/src/components/game/data/zones.ts
+++ b/apps/cryptothrone/astro-cryptothrone/src/components/game/data/zones.ts
@@ -37,7 +37,7 @@ export const ZONES = {
 		tilesetKey: 'cloud-city-tiles',
 		tilesetUrl: '/assets/map/cloud_tileset.png',
 		tilesetName: 'cloud_tileset',
-		channel: '#cryptothrone',
+		channel: '#general',
 		game: 'cryptothrone',
 	},
 } satisfies Record<string, ZoneDef>;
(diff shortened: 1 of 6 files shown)


h0lybyte:
Follow-up in this PR: reviewed send + receive end-to-end. **Send** verified correct (token in URL, socket-open guard, sender/platform pinned, `to_irc_privmsg` round-trips). **Receive** had a gap — the gateway dropped plain IRC messages (no `[CHAT]` wrapper) via `from_irc_privmsg` returning None, so game clients never saw raw IRC chatter in #general. Added `ChatMessage::from_irc_or_plain` (plain fallback → `Chat`/platform `irc`) wired into both i2c forward + history listener. bevy_chat 9/9, gateway 43/43.

PR #12467, opened 2026-06-14
fix(cryptothrone): never CDN-cache non-2xx (sticky /discord/discord.js 404)

h0lybyte:
## Symptom

`https://cryptothrone.com/discord/discord.js` 404s, so the Discord Activity can't boot — even though the deployed 0.1.30 image **contains** the file (verified by extracting `/app/templates/dist/discord/discord.js`, 3.9MB, from the published image).

## Root cause

Cloudflare cached the 404:

| request | result |
|---|---|
| `/discord/discord.js` | 404, `cf-cache-status: HIT`, `age: 46295` (~12.8h) |
| `/discord/discord.js?v=<ts>` | **200** (cache-bust → origin serves the real file) |
| `/discord/` | 200, `cf-cache-status: DYNAMIC` |

The rust `cache_headers` middleware stamped `public, max-age=86400` on **every** static response, including the 404 fallback. So when the bundle was requested before it rolled out, Cloudflare cached that 404 with a 24h TTL and kept serving it.

## Fix

Gate the long `cache-control` on a 2xx status; non-2xx gets `no-store` so a CDN never pins a transient error. Adds a regression unit test (`cargo test -p axum-cryptothrone` → 32 passed). Bumps to 0.1.31 to ship.

## Manual step (needs CF access)

The already-cached 404 won't clear until its max-age expires (~11h left). **Purge Cloudflare cache for `/discord/discord.js`** (and `/embed/embed.js` if cached) to fix it immediately.

### DIFF
--- a/apps/cryptothrone/astro-cryptothrone/vite.discord.config.mts
+++ b/apps/cryptothrone/astro-cryptothrone/vite.discord.config.mts
@@ -35,6 +35,12 @@ export default defineConfig({
 	publicDir: false,
 	define: {
 		'process.env.NODE_ENV': JSON.stringify('production'),
+		// Raw `vite build` (not Astro) does not surface PUBLIC_* from process.env
+		// into import.meta.env, so bake the Discord client id in explicitly from
+		// the Docker build ENV. Public value — safe to inline client-side.
+		'import.meta.env.PUBLIC_DISCORD_CLIENT_ID': JSON.stringify(
+			process.env.PUBLIC_DISCORD_CLIENT_ID ?? '',
+		),
 	},
 	build: {
 		outDir: 'public/discord',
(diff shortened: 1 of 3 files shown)


h0lybyte:
Added a second commit: HTML + the unhashed entry bundles (`discord.js`/`embed.js`) now serve `no-cache` so they revalidate via ETag/304 — fresh on redeploy, cheap when unchanged. Only `/_astro/*` (content-hashed) stays `immutable`. This covers the broader 'stale after deploy' concern, not just the sticky 404. `cargo test -p axum-cryptothrone` → 34 passed.

h0lybyte:
Third commit (same 0.1.31 release): **bake `PUBLIC_DISCORD_CLIENT_ID` into the discord bundle**. After the cache cleared, the Activity failed at runtime with `missing PUBLIC_DISCORD_CLIENT_ID at build time` — the bundle is a raw `vite build`, and Vite (unlike Astro) doesn't surface `PUBLIC_*` from process.env into `import.meta.env`. The Dockerfile `ENV` was set but never reached the bundle. Verified the shipped 0.1.30 `discord.js` contains **0** occurrences of the id; added a `define` that inlines it, rebuilt bundle now embeds it.

PR #12538, opened 2026-06-15
feat(jobboard): dedicated least-privilege DB role to bound the service

h0lybyte:
## Why
RLS **cannot** bound `service_role` (it has `BYPASSRLS`) or `postgres` (superuser) — writing policies for them is a no-op. To actually constrain the jobboard Axum service, it connects as a **dedicated non-owner `jobboard` role** instead (the house forgejo/ows pattern).

## What
- `20260615130000_jobboard_dedicated_role.sql` — create role + `USAGE` + **command-scoped table grants** that encode invariants the DB can hold:
  - **no `DELETE`** on historical/moderation tables (gigs, engagements, applications, reviews, messages, reports)
  - `reviews`/`messages` **immutable** (no `UPDATE`)
  - `audit_log` **append-only** (INSERT/SELECT)
  - lookups read-mostly
  - permissive `for all to jobboard` policy per table so the non-owner role passes RLS/`force`; the GRANTs are the real limit. `execute` on `gen_ulid`/`set_updated_at`; `public`/`extensions` usage. `down` = `drop owned by` + drop role.
- `init/00-roles.sql` — add `jobboard` for local migration testing (parity with `forgejo`).
- `schema/jobboard/jobboard_role.sql` — source-of-truth (generated from the migration).
- Schema stays owned by `postgres` (dbmate does DDL); `jobboard` is DML-only — so grants + RLS genuinely constrain it.

## Verified (PG17 stub)
As the `jobboard` role: INSERT/SELECT work (incl. `gen_ulid` default); `DELETE audit_log/engagements/gigs` and `UPDATE audit_log/messages` are `permission denied`; `postgres` (admin) can still delete. Role up+down cycles cleanly; tables intact.

## Follow-up (infra)
Provision the prod password via a `jobboard-db` sealed secret and point `KBVE_PG_RW_URL` at `postgres://jobboard:…` when the service is deployed.


h0lybyte:
Closing — keeping the existing postgres role via jedi PgCluster (matches the other Axum services). No dedicated jobboard role; postgres is superuser so it bypasses force RLS, and the merged schema (#12507) works as-is.

PR #12553, opened 2026-06-15
fix(jobboard): has_test=false so the docker publish gate isn't blocked

h0lybyte:
Follow-up to merged #12540.

jobboard has no docker e2e harness, and `has_test=true` makes the `test` job a **required gate** before `publish` in ci-docker.yml — with no e2e it would fail and block the ghcr push. The container also can't boot a standalone CI e2e (`PgCluster::from_env` needs a reachable Postgres).

Set `has_test=false` (MDX frontmatter + ci-dispatch-manifest entry) → e2e skipped, `publish` full-rebuilds and pushes `ghcr.io/kbve/jobboard:{version,latest}`. Cargo unit tests still run in the Validate Dev→Main CI. A boot+`/health` docker e2e (needs a throwaway PG, like edge-e2e) can flip it back later.

Verified: `docker build -f apps/jobboard/Dockerfile` succeeds locally — the jobboard binary packages into the chiseled runtime image, so the publish path is sound.


h0lybyte:
Closing — branch is stale (cut before later dev work landed); merging would revert ~9.6k lines. Redoing the lone intended change (has_test=false) cleanly off current dev.

PR #12563, opened 2026-06-16
feat(cryptothrone): fast-register Discord users + fix chat WS jwt (iat)

### LINKED ISSUE #12555: cryptothrone: fast-register Discord users (link-by-email, else create)
## Why
The Discord Activity hands the backend a **verified Discord user**, but `discord.rs` returns **403 "not linked"** when there's no existing KBVE account — a dead end. Provision instead, so launching the Activity is enough to get a playable KBVE session.

Builds on #12552 (the session 500 fix) — land + confirm the existing-linked-user happy path first.

## Strategy (chosen): link-by-email, else create
On session exchange, after the existing-discord-link lookup misses:
1. If Discord email is **verified**, find a KBVE account by that email → **link** the `discord` identity to it.
2. Else → **create** a new Supabase user from the Discord identity.
3. Mint `{jwt, username}` and continue (same as the linked path).

## Required changes
- **Client** (`discord.tsx`): add `email` scope → `authorize({ scope: ['identify','email'] })`.
- **Backend** (`discord.rs`): `/users/@me` already returns `email` + `verified`; pass them down. Replace the 403 branch with a provision-or-link call.
- **DB**: new `SECURITY DEFINER` fn **OWNED BY postgres** (per the 42501 rule — see #12552), e.g. `tracker.ensure_identity_from_discord(p_discord_id, p_username, p_email, p_email_verified)` that does find→link→create in one round-trip and returns `user_id, kbve_username`. (KBVE already manipulates `auth.*` via postgres-owned definer fns; alternative is the GoTrue Admin API, which needs SUPABASE_URL + service-role key wired — decide in design.)
- `profile.username`: derive from the Discord username, sanitized + collision-suffixed; canonical per [[project_supabase_kbve_username_hook]].

## Security guards (must-have)
- **Only link by email when Discord `verified == true`** — never link an unverified email (account-takeover vector).
- Idempotent under concurrent first-launches (unique constraint on discord provider_id; ON CONFLICT do-nothing then re-select).
- New auth users are OAuth-only (no password); confirm that's acceptable vs GoTrue admin provisioning.
- The new fn stays service_role-EXECUTE only, postgres-owned, empty search_path (mirror existing tracker fns).

## Open design decisions
- SQL-definer provisioning vs GoTrue Admin API (env/key wiring).
- Username scheme + collision policy.
- What happens if the matched email account already has a *different* Discord linked.

## Acceptance
Launching the Activity as a Discord user with no prior KBVE link results in a playable session: existing email → linked; no email match → new account created; unverified email → still create (don't link). No 403 dead end.

h0lybyte:
## What
The Discord Activity hands the backend a verified Discord user, so `discord.rs` now **provisions a KBVE account** instead of returning 403 when there's no linked profile.

## Flow
1. `find_claim_identity_by_discord_id` hit → existing path (unchanged).
2. Else **GoTrue admin create-or-get** keyed on email: the verified Discord email, or a stable synthetic `discord_<snowflake>@users.cryptothrone.com` when email is absent/unverified (keeps it idempotent + lets a verified email link to an existing KBVE account). GoTrue owns `auth.users` — **no hand-crafted auth rows**; on "already registered" we resolve via `tracker.find_user_id_by_email`.
3. `tracker.ensure_discord_username` sets `profile.username` (derived from the Discord display name, collision-suffixed via the existing `profile.service_add_username`).
4. Mint the session JWT (unchanged).

## Pieces
- **client** (`discord.tsx`): `scope: ['identify','email']`.
- **SQL** migration `20260616000000`: `tracker.find_user_id_by_email` + `tracker.ensure_discord_username`, both `SECURITY DEFINER` **OWNER postgres** (read-only on `auth.*`; 42501 rule), EXECUTE to service_role.
- **backend** (`discord.rs`): GoTrue admin create-or-get + provision orchestration; uses existing `SUPABASE_URL` + `SUPABASE_SERVICE_ROLE_KEY` env.
- bump **0.1.37**.

## Validation
axum-cryptothrone lint ✓, build-discord ✓ (`scope:["identify","email"]` in bundle). The SQL migration + the GoTrue admin call can only be fully exercised against real Supabase — needs the gated dbmate prod deploy (the two new fns) **and** validation in prod after the image ships. Hand-crafted auth rows were deliberately avoided to de-risk.

## Security
- Link-by-email only via the email key; synthetic email for unverified/no-email so an unverified address never collides onto a real account.
- New SQL fns never write `auth.*` (only GoTrue does); they read auth.users + set profile.username.

Closes #12555.

### DIFF
--- a/apps/cryptothrone/astro-cryptothrone/src/embed/discord.tsx
+++ b/apps/cryptothrone/astro-cryptothrone/src/embed/discord.tsx
@@ -100,7 +100,7 @@ async function boot(): Promise<void> {
 			response_type: 'code',
 			state: '',
 			prompt: 'none',
-			scope: ['identify'],
+			scope: ['identify', 'email'],
 		}),
 	);
 
--- a/apps/cryptothrone/axum-cryptothrone/src/discord.rs
+++ b/apps/cryptothrone/axum-cryptothrone/src/discord.rs
@@ -67,6 +67,7 @@ struct GoTrueUser {
 #[derive(Serialize)]
 struct MintedClaims {
     sub: String,
+    iat: i64,
     exp: i64,
     kbve_username: String,
     role: String,
@@ -333,6 +334,7 @@ fn mint_session_jwt(
 ) -> Result<String, jsonwebtoken::errors::Error> {
     let claims = MintedClaims {
         sub: sub.to_string(),
+        iat: now_secs() as i64,
         exp,
         kbve_username: kbve_username.to_string(),
         role: "authenticated".into(),
--- a/apps/cryptothrone/axum-cryptothrone/src/discord.rs
+++ b/apps/cryptothrone/axum-cryptothrone/src/discord.rs
@@ -17,7 +17,7 @@ use axum::{
     http::StatusCode,
     response::{IntoResponse, Response},
 };
-use jedi::state::pg::{PgCluster, tokio_postgres::Row};
+use jedi::state::pg::{PgCluster, PgConn};
 use jsonwebtoken::{Algorithm, EncodingKey, Header, encode};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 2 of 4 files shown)


h0lybyte:
Added a second fix on this branch: the Activity session JWT minted in `discord.rs` was missing the `iat` claim. irc-gateway's `validate_token` requires `iat` (`set_required_spec_claims [sub,exp,iat,aud]`), so `chat.kbve.com/gamechat` 401'd the token while the game server (no `iat` requirement) accepted it — that's why chat failed only in the Activity (real Supabase JWTs carry `iat`, so the normal build worked). Added `iat=now` to `MintedClaims`. Ships with this PR under 0.1.37.

PR #12759, opened 2026-06-18
feat(unreal): vendor Agones SDK + KBVEAgones ROWS bridge (server-only)

h0lybyte:
## Why

rentearth's UE5 dedicated server has **zero** Agones integration. To run under an Agones Fleet (like cryptothrone-server / nd-server, which call the Agones SDK from the binary), the GameServer must signal `Ready()`/`Health()` — otherwise it never becomes allocatable and crashloops on the health threshold. This adds the missing UE-side lifecycle, kept cleanly separate from KBVEROWS.

## Design (discussed)

- **ROWSPlugin stays a pure REST plugin.** Agones is server-only; coupling it into KBVEROWS would drag an orchestration dep into every client/editor build. Anti-drift comes from vendoring a pinned upstream mirror, not from coupling.
- Lifecycle split mirrors the rest of the repo: **Agones owns Ready/Health**; **ROWS reconciles** allocation + shutdown via its K8s watcher.

## What's here

**Vendored upstream Agones** → `packages/unreal/Agones` (googleforgames/agones `@bb0fb6d`, v2.0.0, Apache-2.0, headers preserved). `UAgonesSubsystem` auto-connects (Ready) + health-pings and self-gates to dedicated server. One deviation: `TargetAllowList: ["Server"]` to keep it out of client compiles — documented in `VENDORING.md`.

**New `KBVEAgones` plugin** (server-only) — the ROWS bridge:
- `UKBVEAgonesBridgeSubsystem` binds the Agones `ConnectedDelegate` (fires post-Ready with GameServer identity) → calls `UROWSInstanceSubsystem::RegisterLauncher(Address, Port, MaxInstances)` using the address/port Agones reports.
- Module `TargetAllowList: ["Server"]` + `ShouldCreateSubsystem` gates on `IsRunningDedicatedServer()`.
- Depends on `Agones` + `ROWS`; KBVEROWS unchanged.
- Config: `GamePortName` (empty = first port), `MaxInstances`.

**rentearth.uproject** — enable `Agones` + `KBVEAgones`.

## Caveats
- **Not build-verified** — no UE toolchain locally; JSON + structure validated only. Real compile needs the UE5 build pipeline.
- rentearth is still a plain K8s **Deployment** — the plugin does nothing live until rentearth is migrated to an Agones Fleet (separate change: Fleet manifest + server Dockerfile/entrypoint).
- Shutdown is left to Agones health-loss + ROWS watcher reconcile; explicit `Shutdown()` can be added if deterministic drain is needed.


h0lybyte:
### Added: 1.58.0 alignment

Two follow-up commits beyond the original plugin scaffold:

- `chore(unreal): pin vendored Agones to release-1.58.0` — re-pin the mirror from a floating master commit to the `release-1.58.0` tag (`ca71cbe`). UE SDK sources are **byte-identical** 1.57→1.58, so doc/pin only.
- `chore(agones): bump cluster Helm chart 1.56.0 → 1.58.0` — `apps/kube/agones/application.yaml` `targetRevision` 1.56.0 → 1.58.0 so the in-cluster control plane matches the vendored SDK. `values.yaml` pins no image tags/feature gates.

**Ops note:** the chart bump rolls the Agones controller/extensions/allocator and applies updated CRDs (ArgoCD ServerSideApply). Two-minor jump (1.56→1.58); running GameServers survive controller restarts. Reaches the cluster only via the normal dev→main release.

Issue #11930, opened 2026-06-07
ops: Forgejo LFS reclaim — deleted stale 'moore' branch, recovered ~22.7 GB


Fudster:
## Summary

Reclaimed **~22.7 GB** on the self-hosted Forgejo (`git.kbve.com`) `kbve/chuck` repo.
The data volume was at 47% (22.9 G / 48.9 G); almost the entire footprint was Git LFS
objects belonging only to a stale `moore` branch. After deleting the branch and running
GC, the volume sits at **0% (189 M used, 48.7 G free)**.

| Metric | Before | After |
|---|---|---|
| `/data/lfs` | 22.9 G | 187 M |
| `/data` used | 22.9 G (47%) | 189 M (0%) |
| chuck branches | `main`, `moore` | `main` |

## Environment

- Cluster: Talos k8s · namespace `forgejo` · Forgejo 12.0.4 (Gitea 1.22 base)
- Storage: PVC `gitea-shared-storage`, 50 Gi Longhorn
- LFS store is shared by `kbve/chuck` + `kbve/rareicon` (rareicon untouched; its ~142 MiB remains)

## Actions taken

1. **Safety snapshot** — Longhorn snapshot `pre-moore-cleanup-20260606` (Ready before proceeding).
2. **Baseline** — `du -sh /data/lfs /data/git` → 22.9 G / 17.3 M; `df -h /data` → 47%.
3. **Deleted `moore`** via Forgejo API (DB-consistent):
   ```
   curl -X DELETE -H "Authorization: token ***" \
     http://localhost:3000/api/v1/repos/kbve/chuck/branches/moore   # 204
   ```
4. **Pruned git** — `git --git-dir=.../kbve/chuck.git gc --prune=now` (17.3 M → 1.8 M).
5. **LFS GC** — `POST /api/v1/admin/cron/gc_lfs`; `/data/lfs` 22.9 G → 187 M on the first pass.
6. **Verified (read-only)** — `forgejo doctor check --run storage-lfs --run check-db-consistency`
   → 785 LFS files (142 MiB), DB **OK**; only `main` remains.

## Follow-ups

- [ ] Revoke 3 short-lived admin tokens (Forgejo → Settings → Applications):
      `moore-cleanup-20260606`, `revoker-tmp-20260606`, `lfsgc-tmp-20260606`
      (token-DELETE API requires basic auth → manual/UI revoke).
[...]

PR #12908, opened 2026-06-21
feat(forgejo): automate LFS GC + Longhorn space reclaim

### LINKED ISSUE #12793: ops: Automate Forgejo LFS GC + Longhorn space reclaim
## Summary

Automate Git LFS garbage collection + space reclaim for the self-hosted Forgejo (`git.kbve.com`) so stale LFS objects don't silently accumulate on the 50 Gi Longhorn PVC. Follow-up from #11930 (manual `moore`-branch cleanup reclaimed ~22.7 GB).

## Background

#11930 showed two distinct layers have to be addressed for space to actually come back:

1. **Forgejo LFS GC** — deletes orphaned LFS objects from the store. Currently only runs on-demand (`POST /api/v1/admin/cron/gc_lfs`). There is no scheduled `[cron.gc_lfs]`.
2. **Block-level reclaim (Longhorn)** — *deleting LFS files in ext4 does not UNMAP the blocks*, so freed space stays pinned in the Longhorn volume/snapshot chain until a **filesystem trim** runs. In #11930, GC alone left ~18.5 GiB stuck in a snapshot; a Longhorn server-side trim (with `unmapMarkSnapChainRemoved=enabled`) was required to actually reclaim it (24.0 GiB → 6.57 GiB actualSize).

Both layers need to be automated, or GC will "succeed" while disk usage stays flat.

## Scope

- [ ] Enable Forgejo scheduled GC in config:
  ```ini
  [cron.gc_lfs]
  ENABLED = true
  SCHEDULE = @every 168h   # weekly; tune as needed
  OLDER_THAN = 168h
  ```
  (Set in the Forgejo values/ConfigMap, not by hand on the pod.)
- [ ] Automate Longhorn-side reclaim for PVC `gitea-shared-storage` (`pvc-0d5edbf3-…`):
  - Either set `unmapMarkSnapChainRemoved: enabled` persistently on the volume, **or**
  - Add a recurring filesystem-trim (Longhorn `RecurringJob` of task `filesystem-trim`, or a CronJob hitting the Longhorn API `?action=trimFilesystem`).
- [ ] Optional: a recurring Longhorn snapshot/backup policy so cleanups always have a safety point without manual snapshots.
- [ ] Document the runbook (GC → trim → verify `actualSize`).

## Environment

- Cluster: Talos k8s · namespace `forgejo` · Forgejo 12.0.4 (Gitea 1.22 base)
- Storage: PVC `gitea-shared-storage`, 50 Gi Longhorn (`pvc-0d5edbf3-53e5-4b1f-8cf2-b28359844f73`)
- LFS store shared by `kbve/chuck` + `kbve/rareicon`

## Verify

`forgejo doctor check --run storage-lfs`, and Longhorn `volume.status.actualSize` should track the live filesystem size after a GC + trim cycle.

Infra work for [Chuck RPG](https://chuckrpg.com).

Fudster:
Closes #12793. Follow-up from #11930.

Stale Forgejo LFS objects accumulate on the 50Gi Longhorn PVC, and reclaiming disk needs **two layers** — deleting LFS objects (ext4 frees inodes only) and then a filesystem trim to unmap the freed blocks from the volume. Without both, GC "succeeds" while disk usage stays flat. PR #12892's `git_gc_repos` cron is git repo GC, not LFS, and does nothing for the block-reclaim layer.

## Changes

**Forgejo** (`apps/kube/forgejo/application.yaml`)
- Add scheduled `[cron.gc_lfs]` (`@every 168h`, `OLDER_THAN = 168h`) alongside the existing `git_gc_repos`; bump `config-revision`.

**Longhorn** (`apps/kube/storage/manifests/`)
- New `recurring-jobs.yaml`: `forgejo-filesystem-trim-weekly` (Sun 03:00) + `forgejo-snapshot-weekly` (Sat 02:00, retain 2), scoped to a dedicated **`forgejo-reclaim` group**.
- `removeSnapshotsDuringFilesystemTrim: true` in `values.yaml` — global equivalent of per-volume `unmapMarkSnapChainRemoved=enabled`, so trim also reclaims space trapped in the snapshot chain (the ~18.5 GiB stuck case in #11930).
- Wire `recurring-jobs.yaml` into the kustomization.

**Docs** (`apps/kube/forgejo/README.md`)
- GC → trim → verify `actualSize` runbook + the volume-label binding step.

## Verified against the live cluster (`admin@kbve`, Longhorn v1.9.1)
- RecurringJob CRD **v1beta2** is the served/storage version → manifest uses it.
- Setting key **`remove-snapshots-during-filesystem-trim` exists** (was `false`) → `values.yaml` key is valid.
- PVC `gitea-shared-storage` → volume `pvc-0d5edbf3-…` (matches the issue), actualSize ~9.2 GiB.
- `kubectl kustomize apps/kube/storage/manifests` renders both RecurringJobs.
- **Scoping label applied** to the forgejo volume: `recurring-job-group.longhorn.io/forgejo-reclaim=enabled`.

## Design call — scoped, not cluster-wide
Longhorn auto-labels **every** volume into the built-in `default` group (confirmed: ~46 volumes, incl. 200–322 GiB DB volumes, several near capacity). A `default`-group snapshot/trim would fan out cluster-wide — a weekly snapshot of all DB volumes is real overhead and risk, counterproductive for a space-reclaim ticket. So the jobs target a dedicated `forgejo-reclaim` group bound only to the LFS volume.

**Trade-off (hidden state):** the group binding is a label on the CSI-created Volume CR, so it lives outside git and must be re-applied if the PVC is recreated. There's no clean GitOps-native way to scope a recurring job to one dynamically-named volume without minting a dedicated StorageClass (which would require recreating the 50Gi PVC). Documented in `recurring-jobs.yaml` + the README runbook; the label is already applied on the live cluster.

## Deferred scope

Cluster-wide `filesystem-trim` (other ~45 volumes incl 200–322Gi DBs) is intentionally **out of scope** here — tracked in #12915. This PR proves the GC→trim→reclaim loop on the single Forgejo LFS volume first to avoid I/O risk on database volumes.

### DIFF
--- a/apps/kube/forgejo/README.md
+++ b/apps/kube/forgejo/README.md
@@ -74,6 +74,64 @@ Forgejo Actions runner (builds with code + assets)
 - **Database**: `forgejo` schema in kilobase CNPG cluster (shared Postgres)
 - **SSH**: NodePort 30022 → `forgejo.kbve.com:30022`
 
+## LFS space reclaim (GC + trim)
+
+Reclaiming disk after stale LFS objects accumulate is a **two-layer** problem
+(see issues #11930 and #12793). Doing only one layer makes GC "succeed" while
+disk usage stays flat:
+
+1. **Forgejo LFS GC** — deletes unreferenced LFS objects from the store
+   (`/data/lfs`). Frees ext4 files/inodes only.
+2. **Longhorn filesystem trim** — unmaps the freed blocks from the volume so
+   `volume.status.actualSize` actually drops. ext4 deletes do not UNMAP blocks,
+   so without a trim the space stays pinned in the Longhorn volume/snapshot
+   chain.
+
+### Automation (current state)
+
+- **GC**: `[cron.gc_lfs]` runs weekly (`@every 168h`, `OLDER_THAN = 168h`),
+  configured in `apps/kube/forgejo/application.yaml`.
+- **Trim**: Longhorn `RecurringJob forgejo-filesystem-trim-weekly` (Sun 03:00) in
+  `apps/kube/storage/manifests/recurring-jobs.yaml`, with
+  `removeSnapshotsDuringFilesystemTrim: true` (values.yaml) so snapshot-chain
+  space is reclaimed too. A `forgejo-snapshot-weekly` job (Sat 02:00, retain 2)
+  provides a safety point before the trim.
+- **Scoping**: both jobs target the `forgejo-reclaim` group, **not** Longhorn's
+  cluster-wide `default` group, so they only touch the LFS volume. Binding is via
+  a label on the CSI-created Volume CR (not in git) — re-apply it if the PVC is
+  ever recreated:
+
+  ```bash
+  kubectl -n longhorn-system label volume \
+    $(kubectl -n forgejo get pvc gitea-shared-storage -o jsonpath='{.spec.volumeName}') \
+    recurring-job-group.longhorn.io/forgejo-reclaim=enabled --overwrite
+  ```
+
+### Manual runbook (on-demand reclaim / verification)
+
+```bash
+# 1. Run LFS GC now (instead of waiting for the weekly cron).
+#    Forgejo admin API — token must have admin scope.
+curl -X POST -H "Authorization: token <ADMIN_TOKEN>" \
+  https://git.kbve.com/api/v1/admin/cron/gc_lfs
+
+# 2. Confirm the store shrank inside the pod.
+kubectl exec -n forgejo deploy/forgejo -- du -sh /data/lfs
+
+# 3. Trigger the Longhorn filesystem trim for the volume (or run the
+#    RecurringJob ad-hoc). Easiest via the Longhorn UI:
+#    Volume → gitea-shared-storage → "Trim Filesystem".
+#    API equivalent:
+#    POST /v1/volumes/<vol>?action=trimFilesystem
+
+# 4. Verify reclaim — actualSize should track the live filesystem size.
+kubectl get volumes.longhorn.io -n longhorn-system \
+  -o custom-columns=NAME:.metadata.name,SIZE:.spec.size,ACTUAL:.status.actualSize
+
+# 5. Storage-layer health.
+kubectl exec -n forgejo deploy/forgejo -- forgejo doctor check --run storage-lfs
+```
+
 ## DNS
 
 Add `forgejo.kbve.com` A record pointing to `142.132.206.74` in Cloudflare.
--- a/apps/kube/forgejo/README.md
+++ b/apps/kube/forgejo/README.md
@@ -89,8 +89,9 @@ disk usage stays flat:
 
 ### Automation (current state)
 
-- **GC**: `[cron.gc_lfs]` runs weekly (`@every 168h`, `OLDER_THAN = 168h`),
-  configured in `apps/kube/forgejo/application.yaml`.
+- **GC**: `[cron.gc_lfs]` (`SCHEDULE = '0 1 * * 0'`, `OLDER_THAN = 168h`),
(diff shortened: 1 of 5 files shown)


h0lybyte:
Audited the PR. Two-layer GC→trim model is sound and the LFS-vs-block distinction is correct. A few things worth tightening before merge:

### 1. `default` group is cluster-wide, not forgejo-scoped — biggest concern
Both RecurringJobs target the built-in `default` group, which Longhorn auto-applies to **every** volume lacking an explicit recurring-job label. There are currently **no** recurring jobs in the cluster (`values.yaml` only sets `backupstorePollInterval`/`disableSnapshotPurge`), so this PR introduces the first-ever cluster-wide:

- **weekly snapshot (retain 2) on every Longhorn volume** — supabase-cluster, valkey, the 300Gi engine cache, etc., not just the 50Gi forgejo PVC. That's new snapshot-storage overhead and new I/O on DB volumes, framed in the PR as a forgejo safety point.
- **weekly filesystem-trim on every volume** + **global `removeSnapshotsDuringFilesystemTrim: true`**, which alters the snapshot-chain behavior of all those DB volumes too.

The PR body justifies `default` as "simpler than minting a StorageClass," but the trade-off (cluster-wide blast radius vs. one PVC) isn't surfaced. If only forgejo needs this, scope it with a `recurringJobSelector` label on the forgejo volume + a named group instead of `default`. If cluster-wide snapshots are actually desired, call that out explicitly — it's a meaningful policy change.

### 2. GC and trim aren't actually ordered
The trim job is pinned to `Sun 03:00` "after the weekly Forgejo gc_lfs run," but `gc_lfs` uses `SCHEDULE: '@every 168h'` with `RUN_AT_START: false` — that's a rolling 7-day timer anchored to pod start, **not** a Sunday cron. So gc can fire any day/time and the "trim runs after gc" ordering the comments claim isn't guaranteed. Either pin gc with a cron expression (e.g. `SCHEDULE: '@weekly'` or `0 1 * * 0`) so it precedes the Sun 03:00 trim, or drop the ordering claim and rely on eventual reclaim (next trim picks up whatever the last gc freed — which is fine, just not what the comments say).

### 3. Minor
- `snapshot-weekly` Sat 02:00 + `filesystem-trim` Sun 03:00 leaves the safety snapshot ~25h stale by trim time — acceptable, but if a bad gc lands Sunday morning the snapshot predates it. Tighter spacing (or snapshot immediately pre-trim) gives a truer rollback point.
- Confirm `config-revision` bump alone reliably rolls the pod (Reloader/annotation), since the new `[cron.gc_lfs]` block only takes effect on restart.

Nothing blocking the LFS GC logic itself — concern is the unscoped Longhorn blast radius (#1) and the false ordering guarantee (#2).

h0lybyte:
Pushed two follow-up commits for the remaining review items (#1 scoping was already handled by your `forgejo-reclaim` group change — kept as-is):

- **GC/trim ordering** — `gc_lfs` was `@every 168h` (anchored to pod start, so no guaranteed ordering vs the Sun 03:00 trim). Pinned to `SCHEDULE: '0 1 * * 0'` (Sun 01:00) so it deterministically precedes the trim. README updated.
- **Pod roll** — `config-revision` was under `deployment.annotations` (Deployment *metadata* → no pod-template hash change → no rollout). Moved to `podAnnotations` so bumping it actually rolls the pod and loads the new `[cron.gc_lfs]` config.

`kubectl kustomize apps/kube/storage/manifests` still renders clean.

h0lybyte:
## Follow-ups (post-merge)

- #12920 — activation (apply `forgejo-reclaim` label after deploy) + long-term declarative binding via StorageClass `recurringJobSelector` to retire manual labels.
- #12915 — extend `filesystem-trim` to other Longhorn volumes (group-by-group, I/O testing).

This PR stays scoped to the single Forgejo LFS volume.

PR #13318, opened 2026-06-25
chore(arc): retire orphaned arc-runner-ue-mac scale set

h0lybyte:
## Context
Now that Mac UE builds run on the self-hosted **MacBook Air** (`runs-on: [self-hosted, macOS, ARM64]`, registered directly with GitHub), the `arc-runner-ue-mac` **ARC scale set** is dead scaffolding — it required a macOS k8s node (`nodeSelector: kbve.com/os=macos`) that no longer exists, so it sat at **0 runners** with an idle listener pod in `arc-systems`, and **no workflow targets its label**.

## Audit (done before removal)
| Check | Result |
|---|---|
| `runs-on: arc-runner-ue-mac` anywhere | **0 matches** across all workflows |
| Argo runners app sync policy | `prune: true` + `selfHeal: true` → scaleset + listener **auto-pruned** on sync (no orphan) |
| Scale set finalizer | present but `ephemeral=0` → clears cleanly (the known wedge needs active runners + RBAC churn) |
| `github-arc-token` | **ESO-owned** (shared by all 5 runner releases) → not deleted |
| MacBook Air runner | separate label + direct-GitHub registration → **unaffected** |
| No KubeVirt mac VM / PVC / DataVolume | confirmed already gone (#13231) |

## Changes
- Remove the `arc-runner-ue-mac` Helm release from `application.yaml` + delete `values-ue-mac.yaml`
- Drop the `arc-runner-ue-mac` label from `.github/actionlint.yaml`
- Fix 3 stale docs that named `arc-runner-ue-mac` (real target is `[self-hosted, macOS, ARM64]`)

## Notes
- ArgoCD tracks `main` → cluster cleanup (scaleset + listener pod) happens after the next dev→main release sync. If the AutoscalingRunnerSet finalizer ever wedges: `kubectl -n arc-runners patch autoscalingrunnerset arc-runner-ue-mac --type=merge -p '{"metadata":{"finalizers":null}}'`.
- Follow-up: the `UE5-Mac` actionlint label also appears orphaned (only a job *name*, never a `runs-on`) — left in place for now.


h0lybyte:
**h0lyMac** _(via Discord)_:

<@​256263372977209355> can you audit this for me

Fudster:
## 🔍 Production Audit — Approve

Ran this against a production-grade PR audit framework (11 dimensions). Independently verified every claim in the PR's audit table against the tree, including the non-obvious ones.

### Verified
| Claim | Result |
|---|---|
| No `runs-on: arc-runner-ue-mac` anywhere | ✅ 0 references post-PR |
| Deleting `values-ue-mac.yaml` is safe | ✅ **confirmed** — `manifests/kustomization.yaml` never lists any `values-*.yaml`, so the kustomize directory source ignores them entirely |
| `$values` ref / source ordering unaffected | ✅ `ref: values` source unchanged; ArgoCD resolves `$values` by ref name, not position |
| `github-arc-token` ESO-owned, shared | ✅ still referenced by the other 4 releases, not deleted |
| Other 4 runner releases independent | ✅ distinct `releaseName`s, no shared rendered objects |

### Risk surface
- **Net reduction** — removes an idle listener pod (50m/64Mi) and a runner that ran `RUNNER_ALLOW_RUNASROOT=1`. No RBAC/secret/endpoint exposure introduced.
- **Rollback:** revert re-adds the source; chart pinned `0.9.3` → clean.
- **Deployment timing:** ArgoCD tracks `main`, so cluster prune happens on the next dev→main sync, not at merge.

### Findings
**M1 (Medium) — finalizer wedge depends on a live controller at prune time.** The `AutoscalingRunnerSet` cleanup finalizer clears cleanly only while `arc-controller` is up. With 0 runners the known wedge is avoided (as the PR notes), and the `kubectl patch ... finalizers:null` fallback is documented — already mitigated. Worst case is a transient `arc-runners` OutOfSync during the release sync, cleared in seconds.

**L1** — pre-existing: `manifests/runner-networkpolicy.yaml` is on disk but absent from `kustomization.yaml`, so it's never applied. Unrelated to this PR; worth a follow-up.

**L2** — deferred cleanup + finalizer fallback live only in the PR description; a one-line runbook note would help it survive past merge.

**L3** — `UE5-Mac` actionlint label now also orphaned (already noted as follow-up).

### Decision: **APPROVE** ✅
Accurate, contained, reversible. No blockers.

**Most likely 30-day incident:** a transient `arc-runners` OutOfSync during the dev→main sync if the `AutoscalingRunnerSet` hangs on its finalizer while `arc-controller` restarts — fixed in seconds by the documented patch. The live MacBook Air runner is on a separate label/registration path and is fully unaffected.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

PR #13707, opened 2026-07-02
fix(KBVEYYJson): verify packaged yyjson.h include resolution across platforms

### LINKED ISSUE #13493: sync(KBVEYYJson): fix 'yyjson.h not found' breaking KBVEItemDB build
Part of the Chuck ↔ monorepo plugin sync (#13464). **Blocker** — surfaced by the KBVESQLite PR (#13471): `KBVEItemDB` Mac verify fails on a KBVEYYJson packaging bug, not on the changed plugin.

## Failure
```
KBVEYYJson.h:7:10: fatal error: 'yyjson.h' file not found
```
Under `BuildPlugin -StrictIncludes -NoPCH -DisableUnity`, `KBVEYYJson.h` (public) `#include`s the third-party `yyjson.h`, but the packaged plugin build can't resolve the path. Any consumer (KBVEItemDB, and its `Generated/KBVEItemDBProtoParse.h`) fails to compile.

## Scope
- Make `yyjson.h` resolvable from `KBVEYYJson`'s **public** include path (ThirdParty include dir added as `PublicIncludePaths`, or vendor the header into `Public/`), so consumers compile.
- Wrap the third-party include in `THIRD_PARTY_INCLUDES_START/END` if not already.
- Verify on all three platforms via `ci-unreal-plugins.yml`.

## Done when
- `Build Plugin {Mac,Linux,Win64} (KBVEYYJson)` green, and KBVEItemDB compiles against it.

h0lybyte:
Closes #13493.

## What
Normalize the `ThirdParty/yyjson` public include path in `KBVEYYJson.Build.cs` with `Path.GetFullPath`, and drop a stray comment.

## Why
Issue #13493 tracked `fatal error: 'yyjson.h' file not found` when consumers (KBVEItemDB) compiled against the packaged KBVEYYJson plugin under `BuildPlugin -StrictIncludes`. The root packaging bug — `ThirdParty/` missing from the packaged plugin — was fixed by adding `Config/FilterPlugin.ini` (79c04e49e, shipped with #13471). KBVEQuestDB's Mac verify (run 28496953459, PR #13634) already compiles green against packaged KBVEYYJson, including its `Generated/` proto-parse header that includes `KBVEYYJson.h`.

Remaining checklist from the issue:
- `yyjson.h` resolvable from public include path — already done (`PublicIncludePaths` + FilterPlugin.ini); this PR normalizes the path.
- `THIRD_PARTY_INCLUDES_START/END` — already wrapped in `KBVEYYJson.h` and `KBVEYYJsonThirdParty.cpp`.
- Verify all three platforms — this PR carries the `ci-full-plugins` label so `ci-unreal-plugins.yml` runs Mac + Linux + Win64 for KBVEYYJson and its reverse-deps (KBVEItemDB, KBVENPCDB, KBVEQuestDB, UEDevOps).

## Note
KBVESpellDB declares a KBVEYYJson plugin dependency but is absent from `.github/ci-dispatch-manifest.json`, so it gets no plugin verify. Needs a manifest regen (`sync:ci-manifest`) in a follow-up.

### DIFF
--- a/packages/unreal/KBVENPCDB/Source/KBVENPCSprite/Private/KBVENpcSpriteRenderSubsystem.cpp
+++ b/packages/unreal/KBVENPCDB/Source/KBVENPCSprite/Private/KBVENpcSpriteRenderSubsystem.cpp
@@ -4,6 +4,7 @@
 #include "Components/InstancedStaticMeshComponent.h"
 #include "Engine/StaticMesh.h"
 #include "Engine/Texture.h"
+#include "Engine/Texture2D.h"
 #include "Engine/World.h"
 #include "Materials/Material.h"
 #include "Materials/MaterialInstanceDynamic.h"
--- a/packages/unreal/KBVEYYJson/Source/KBVEYYJson/KBVEYYJson.Build.cs
+++ b/packages/unreal/KBVEYYJson/Source/KBVEYYJson/KBVEYYJson.Build.cs
@@ -12,7 +12,7 @@ public class KBVEYYJson : ModuleRules
 			"Core"
 		});
 
-		string ThirdPartyDir = Path.Combine(ModuleDirectory, "..", "..", "ThirdParty", "yyjson");
+		string ThirdPartyDir = Path.GetFullPath(Path.Combine(ModuleDirectory, "..", "..", "ThirdParty", "yyjson"));
 
 		PublicIncludePaths.Add(ThirdPartyDir);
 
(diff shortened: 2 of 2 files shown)


h0lybyte:
Pushed 4088fe155: KBVENPCDB's Build Plugin Linux failed on `KBVENpcSpriteRenderSubsystem.cpp:306` — `UTexture2D*` → `UTexture*` upcast rejected because the TU only saw the forward declaration (no `Engine/Texture2D.h` include) under -StrictIncludes. Latent pre-existing break, surfaced here because KBVENPCDB re-verifies as a KBVEYYJson consumer. One-line include fix.

PR #14108, opened 2026-07-14
feat(vector): warn+/error+ only to ClickHouse across all namespaces

h0lybyte:
## Goal

Cut the INFO firehose into `observability.logs_distributed` — capture warn+/error+ cluster-wide for fast error triage, drop INFO (the bulk of the volume that just hammers the DB).

## Change

`noise_filter` previously dropped INFO/debug for **only 12 named infra namespaces**; every app namespace still shipped INFO. This makes the filter **global**:
- Keep `warn / warning / error / fatal / panic / critical` (parsed `.severity` / `.metadata.level` first, message-text regex fallback for unstructured logs).
- Drop `info / debug / trace / notice` for **every** namespace.

Per-namespace visibility is preserved: the `namespace_heartbeat` branch (1 row/ns/5min, ~600 rows/hr) **bypasses `noise_filter`**, so quiet namespaces with no warn+/error+ still register on the dashboard's namespace count.

## Lossless reformat

Also converts the `vector.yml` ConfigMap from an unmaintainable escaped one-liner into a YAML block scalar. **Proven lossless** — parsing both versions, the content delta is confined entirely to the `noise_filter` block (9 lines out / 15 in); all other sources/transforms/sinks are byte-identical. The +525/-189 line count is the escaping change, not behavior.

## Notes / follow-ups

- Source still excludes `kube-system` + `vector` paths (52 ns total, ~44 tracked). If you want cluster-control-plane errors too, we can drop the `kube-system` source exclude — only its warn+/error+ would ship under the new filter. Say the word.
- Further INFO reduction is now centralized here; source-side app log-level tuning is a separate lever if specific namespaces still dominate warn-level volume.
- No sink/schema change; `skip_unknown_fields: true` unchanged.

📚 https://kbve.com/application/kubernetes/

### DIFF
--- a/apps/kube/vector/manifests/vector-config.yaml
+++ b/apps/kube/vector/manifests/vector-config.yaml
@@ -3,196 +3,532 @@ kind: ConfigMap
 metadata:
     name: vector-config
     namespace: vector
-    annotations:
-        kbve.sh/last-updated: '2026-06-29T00:00:00Z'
-        kbve.sh/config-version: v8-structured-metadata
     labels:
         kbve.com/category: observability
-        kbve.com/stack: core
+        kbve.com/stack: vector
         kbve.com/managed-by: argocd
 data:
-    vector.yml:
-        "api:\n  enabled: true\n  address: 0.0.0.0:9001\n\nsources:\n  kubernetes_logs:\n\
-        \    type: kubernetes_logs\n    self_node_name: \"${VECTOR_SELF_NODE_NAME}\"\n\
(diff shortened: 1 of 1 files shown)


h0lybyte:
Superseded — keeping the broad vector net (catch-all) and instead reducing log volume at the source for the specific apps that over-emit INFO. metallb-system (the 46% info / 97% error dominator) is already removed; remaining work is scoped per-app log-level tuning.

Issue #14423, opened 2026-07-20
Harden pnpm monorepo against npm supply-chain attacks (bwrap sandbox + release-age)


h0lybyte:
## Context

Rise in npm supply-chain attacks (shai-hulud-style worms, compromised-then-yanked releases landing within hours). Hardening our pnpm monorepo against malicious lifecycle/build scripts.

Prior art / discussion: https://news.ycombinator.com/item?id=45034496 — bubblewrap wrapper for isolating package managers, plus `npm config set ignore-scripts true`.

## Already in place (baseline)

- **`allowBuilds` allowlist** in `pnpm-workspace.yaml` — pnpm 11 blocks all lifecycle scripts by default; only 8 explicitly-allowed deps run postinstall. This is the granular equivalent of `ignore-scripts`, and stronger.
- **`--frozen-lockfile`** enforced in CI (`utils-nx-kbve-shell.yml`).
- **`overrides`** pin known-vulnerable transitive deps.

## Done in this branch

Branch `worktree-npm-hardening-bwrap`:

- [x] Bump `minimumReleaseAge` `720` → `2880` in `pnpm-workspace.yaml` — quarantine window on freshly-published versions raised from 12h to 48h. Most npm compromises are caught and yanked inside 24h.
- [x] Add `scripts/ci/bwrap-pnpm.sh` — bubblewrap sandbox wrapper for `pnpm install` on **Linux CI runners only**. Confines FS to workspace + pnpm store (RO-root), tmpfs's `$HOME` to hide credentials, and **scrubs secret env vars** (`--clearenv` + allowlist) so a compromised postinstall cannot read `$GITHUB_TOKEN` / registry tokens. macOS = passthrough (bwrap is Linux-namespaces only).

### 1. Runner image: install bubblewrap
- [ ] Add `bubblewrap` to the ARC/KubeVirt Linux runner image.
- [ ] **Gotcha:** ARC runners are k8s pods; unprivileged user namespaces (`clone(CLONE_NEWUSER)`) are typically blocked by the default seccomp/AppArmor profile. Either ship a **setuid `bwrap`** binary, or set pod `securityContext` to permit userns. Verify which path the runner set allows before wiring.

### 2. Wire the wrapper into CI (after #1)
- [ ] Swap the install step in `.github/workflows/utils-nx-kbve-shell.yml` (currently line ~76):
  ```yaml
  - name: Install pnpm Dependencies
[...]

PR #14430, opened 2026-07-20
feat(ci): bwrap sandbox wrapper for pnpm install (Linux runners)

h0lybyte:
Part of #14423.

Adds `scripts/ci/bwrap-pnpm.sh` — a [bubblewrap](https://github.com/containers/bubblewrap) sandbox wrapper for `pnpm install` on **Linux CI runners**.

### What it does
- **Env scrub (primary defense):** `--clearenv` + an allowlist, so a compromised lifecycle/build script cannot read `$GITHUB_TOKEN`, registry tokens, or other CI secrets.
- **FS confinement:** RO-root, RW holes only for the workspace + pnpm store; `$HOME` is tmpfs'd to hide `~/.ssh` / `~/.aws` / `~/.npmrc` auth.
- **macOS = passthrough** (bwrap is Linux-namespaces only).
- `BWRAP_KEEP_ENV` escape hatch for a private-registry token if ever needed.

### Not wired in yet — intentional
This PR only adds the script. It is **not** referenced by any workflow, so it changes nothing at runtime. Two blockers before wiring (tracked in #14423):
1. `bubblewrap` must be in the ARC/KubeVirt runner image — and ARC pods usually block unprivileged userns, so it needs a **setuid `bwrap`** or a `securityContext` change.
2. Then swap the install step in `utils-nx-kbve-shell.yml` to `./scripts/ci/bwrap-pnpm.sh install --frozen-lockfile`.

### Verification
`bash -n` + `shellcheck` clean.

### DIFF
--- a/scripts/ci/bwrap-pnpm.sh
+++ b/scripts/ci/bwrap-pnpm.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+#
+# bwrap-pnpm.sh — run `pnpm install` (or any pnpm cmd) inside a bubblewrap sandbox
+# on Linux CI runners. Confines the filesystem to the workspace + pnpm store and
+# strips secret env vars, so a malicious lifecycle/build script from a compromised
+# npm package cannot read CI secrets ($GITHUB_TOKEN, registry tokens) or touch the
+# runner filesystem outside the repo.
+#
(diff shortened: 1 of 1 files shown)


h0lybyte:
No longer required, going to close it out.

PR #14661, opened 2026-07-25
feat(reel): fail unresolvable torrents with a reason instead of hanging

h0lybyte:
## Problem

Adding a magnet with no reachable peers (dead swarm, DHT-only `tr=dht://`, unreachable trackers) leaves the torrent stuck in **Leeching forever**. `wait_until_completed()` never returns *and* never errors, so the UI shows a permanent spinner with no explanation. Reproduced live on `kbve.com/media/reel/` — add succeeds, nothing ever happens, no error.

## Fix — backend (`engine.rs` completion watcher)

Two guards before the mover runs:

1. **Metadata gate** — `tokio::time::timeout(REEL_METADATA_TIMEOUT_SECS, handle.wait_until_initialized())`. On timeout/err → `Failed` + *"could not resolve torrent metadata within Ns — no peers responded (dead magnet or unreachable trackers)"*, then purge the partial from the session.
2. **Stall watchdog** — after metadata resolves, `select!` between `wait_until_completed()` and a `REEL_STALL_CHECK_SECS` tick reading `handle.stats().progress_bytes`. No forward progress for `REEL_STALL_TIMEOUT_SECS` → `Failed` + *"no data received for Ns — no seeders have the content"*.

New config (all defaulted, no deploy change required):

| env | default |
|---|---|
| `REEL_METADATA_TIMEOUT_SECS` | 120 |
| `REEL_STALL_TIMEOUT_SECS` | 300 |
| `REEL_STALL_CHECK_SECS` | 15 |

Pure `is_stalled()` helper + unit test. **111 tests pass** (was 110).

## Fix — frontend (`reelService.ts`)

Player now checks torrent state after fetching detail: `Failed` → shows the backend reason; `Reaped` → "expired, re-add to watch". Previously it probed the manifest of a dead torrent and showed a confusing status. The staff console (#14648) already renders the `Failed` badge + error.

## Verify

- `cargo test -p reel`: 111 passed
- `tsc --noEmit`: 0 errors in reel/media files

### DIFF
--- a/apps/kbve/astro-kbve/src/components/media/reelService.ts
+++ b/apps/kbve/astro-kbve/src/components/media/reelService.ts
@@ -168,6 +168,18 @@ export class ReelPlayer {
 			);
 			if (this.generation !== gen) return;
 			$reelName.set(detail?.name ?? null);
+			if (detail?.state === 'Failed') {
+				this.fail(
+					typeof detail.error === 'string' && detail.error
+						? detail.error
+						: 'this reel failed to download',
+				);
+				return;
+			}
+			if (detail?.state === 'Reaped') {
+				this.fail('this reel expired and was removed — re-add it to watch');
+				return;
+			}
 		} catch (e) {
 			if (this.generation !== gen) return;
 			if (e instanceof ApiError && e.status === 404) {
--- a/apps/media/reel/src/config.rs
+++ b/apps/media/reel/src/config.rs
@@ -13,6 +13,9 @@ pub struct Config {
     pub metadata_timeout_secs: u64,
     pub stall_timeout_secs: u64,
     pub stall_check_secs: u64,
+    pub extra_trackers: Vec<String>,
+    pub trackers_url: Option<String>,
+    pub trackers_refresh_secs: u64,
     pub state_flush_ms: u64,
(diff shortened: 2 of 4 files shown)


h0lybyte:
Added a second commit: **public tracker injection + auto-refresh**.

Root of the stalls wasn't just the missing timeout — magnets carrying only `wss://` (webtorrent) or `dht://` trackers had no usable swarm for librqbit. reel now unions a public open-tracker list into every add:

- Embedded default UDP/HTTP list (first-boot + offline fallback)
- `tracker_refresh_loop` pulls [ngosang/trackerslist `trackers_best.txt`](https://github.com/ngosang/trackerslist) from GitHub raw on boot + every `REEL_TRACKERS_REFRESH_SECS` (6h), hot-swapped under a mutex; failed fetch keeps the previous list
- `parse_trackers` filters to udp/http/https (ws/wss unusable by librqbit)
- Config: `REEL_TRACKERS_URL` (none/off disables pull), `REEL_TRACKERS` (manual override), `REEL_TRACKERS_REFRESH_SECS`

Trackers only coordinate peers (announce lists), not content — using a public list is legal; reel's legal-content policy still governs what gets added. 114 tests pass.

PR #14749, opened 2026-07-27
fix(reel): non-blocking magnet add + auto-refreshing console

h0lybyte:
## The Cloudflare 520 on add
librqbit's `add_torrent()` **blocks until it resolves metadata** from peers (session.rs `resolve_magnet(...).await`). Adding a magnet with no live seeders (e.g. a rare public-domain torrent whose only tracker is dead) hangs the call forever → the axum-kbve reel proxy times out → the browser gets *"origin returned an invalid or incomplete response"* (Cloudflare 520), and the console sits on **'Adding…'** indefinitely.

My earlier metadata-timeout (#14661) didn't help: it gated on `wait_until_initialized` *after* `add_torrent` returns — which never happens for an unresolvable magnet.

## Backend fix — return immediately, resolve in background
A magnet already carries its info_hash, so `add()` now:
1. Parses the info_hash via `librqbit::Magnet::parse` (instant).
2. Registers the torrent as `Leeching` and **returns the id right away**.
3. Runs `add_torrent` in a background task bounded by `REEL_METADATA_TIMEOUT_SECS`; on timeout/error the entry flips to `Failed` with a reason.

Non-magnet `.torrent` URLs (info_hash unknown upfront) resolve inline but are now `timeout`-bounded so the request can't hang either.

## Frontend fix — no more frozen 'Adding…'
The staff console now polls every 4s while any torrent is `Leeching` or transcoding/HLS-active, so `Leeching → Seeding/Failed` transitions and failure reasons appear on their own. (Survey gap #6.)

## Verify
- `cargo test -p reel`: 117 passed; clean build.
- `tsc --noEmit`: 0 errors in reel/media files.

Bumps reel.mdx → 0.1.7. Auto-deploys once #14745 (deployment tag pin) lands; otherwise one `kubectl rollout restart` picks it up.

Note: the specific magnet you tried (The Blancheville Monster) will now fail *gracefully* to `Failed: could not resolve metadata` if it genuinely has no seeders — the fix stops the hang, it can't conjure peers for a dead swarm.

🤖 Generated with [KBVE Rust Code](https://kbve.com/application/rust/)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15415,6 +15415,7 @@ dependencies = [
  "tower 0.5.3",
  "tracing",
  "tracing-subscriber",
+ "url",
 ]
 
 [[package]]
--- a/apps/kbve/astro-kbve/src/components/media/ReactReelConsole.tsx
+++ b/apps/kbve/astro-kbve/src/components/media/ReactReelConsole.tsx
@@ -50,6 +50,22 @@ export default function ReactReelConsole() {
 		if (isStaff) void refreshReelList();
 	}, [isStaff]);
 
+	useEffect(() => {
+		if (!isStaff) return;
+		const active = list.some(
+			(t) =>
+				t.state === 'Leeching' ||
+				t.transcode === 'Pending' ||
+				t.transcode === 'Remuxing' ||
+				t.transcode === 'Encoding' ||
+				t.hls === 'Starting' ||
+				t.hls === 'Live',
+		);
+		if (!active) return;
+		const timer = setInterval(() => void refreshReelList(), 4000);
+		return () => clearInterval(timer);
+	}, [isStaff, list]);
+
 	const add = useCallback(async () => {
 		const src = source.trim();
 		if (!src) return;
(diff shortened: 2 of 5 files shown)


h0lybyte:
Added a third commit answering "shouldn't it use our tracker list?" — **it wasn't**, and now does.

librqbit only merges `opts.trackers` (reel's injected list) for **.torrent-file** adds; in the **magnet** branch it uses only the magnet's own trackers and drops `opts.trackers` (session.rs). So the #14661 tracker injection never helped magnet resolution — a magnet with one dead tracker (like your Tears of Steel with only `openbittorrent:80`) couldn't reach the 48 good trackers reel maintains.

Fix: append reel's tracker list to the magnet as percent-encoded `&tr=` params (via `url::Url::query_pairs_mut`) before handing it to librqbit, so they land in `magnet.trackers` and actually drive metadata resolution + peer discovery. + `magnet_with_trackers` unit test (parses the augmented magnet back through librqbit). 118 tests.

PR #14944, opened 2026-07-30
fix(sql): enable + harden RLS on windmill policy tables

h0lybyte:
## Problem

Supabase linter `policy_exists_rls_disabled` (CRITICAL) on 5 relations:

- `windmill.account`
- `windmill.v2_job_completed`
- `windmill.v2_job_queue`
- `windmill.v2_job_runtime`
- `windmill.v2_job_status`

Windmill's sqlx migrations created policies (`admin_policy`) on these tables but left `ROW LEVEL SECURITY` disabled, so the policies never apply.

## Changes

Two new dbmate migrations:

### `20260730020000_windmill_rls_enable_policy_tables.sql` — resolve

One-shot sweep: every relation in schema `windmill` with `relkind IN ('r','p')`, `NOT relrowsecurity`, and at least one `pg_policy` row gets `ALTER TABLE … ENABLE ROW LEVEL SECURITY`. No table names hardcoded — Windmill owns its table lifecycle (`20260707000000_windmill_schema_init` only creates schema, roles, grants). Idempotent; no-op on a fresh database where Windmill has not migrated yet. Verification block raises if any offender remains.

### `20260730030000_windmill_rls_hardening.sql` — harden

The one-shot sweep only holds until Windmill's next sqlx migration adds another policy-bearing table with RLS off, so the invariant is made durable:

- `windmill.rls_drift()` — read-only audit of policy-without-RLS relations (empty result = healthy)
- `windmill.enforce_policy_rls()` — re-runnable sweep, returns number of relations changed
- pg_cron job `windmill-enforce-policy-rls` (hourly at :17) — heals drift after Windmill migrations run; guarded by extension presence, idempotent unschedule-then-schedule
- privilege wall: `REVOKE ALL ON ALL TABLES / SEQUENCES / FUNCTIONS` plus `ALTER DEFAULT PRIVILEGES … REVOKE` from `PUBLIC`, `anon`, `authenticated`, so the API roles cannot reach windmill internals even if a future object ships with permissive grants

Both functions are `SECURITY DEFINER`, owned by `postgres`, `search_path` pinned to `pg_catalog`, `EXECUTE` revoked from `PUBLIC`/`anon`/`authenticated`.

## Why this is safe

Windmill tables are owned by `postgres` and the Windmill server connects as `postgres`, so RLS is bypassed for the engine itself. Workers `SET ROLE` to `windmill_user` / `windmill_admin` (BYPASSRLS) — exactly the enforcement path these policies were written for.

**Escape hatch:** if Windmill ever ships a policy-bearing table that must run *without* RLS, unschedule `windmill-enforce-policy-rls` rather than editing the migration (documented in the migration header).

## Testing

`test-migration.sh` could not be used: the baseline chain dies locally at `schema "auth" does not exist` (known local-kilobase limitation), and the shared `dbmate-postgres-1` container was being recycled by another session mid-run.

Tested isolated instead, against a throwaway `postgres:17-alpine` seeded with `init/00-roles.sql` for the supabase roles:

- both migrations: seed → up → assert_after_up → down → assert_after_down → re-up, all green
- end state verified: policy-bearing probe tables `relrowsecurity=true`, no-policy control tables `false`, `rls_drift()` clean, no `PUBLIC`/`anon`/`authenticated` table grants, both functions `prosecdef=true` and owned by `postgres`, `enforce_policy_rls()` returns `0` on a second call (idempotent)
- cron branch (pg_cron absent in alpine) exercised against `cron.schedule` / `cron.unschedule` stubs with the guard forced: schedules once, re-run creates no duplicate, `migrate:down` removes it

Bug caught and fixed during testing: a single `IF EXISTS (pg_extension …) AND NOT EXISTS (cron.job …)` fails with `relation "cron.job" does not exist` on installs without pg_cron, because plpgsql plans the whole boolean expression — split into nested `IF`s.

Companion `.test.sql` fixtures added for both migrations.

### DIFF
--- a/packages/data/sql/dbmate/migration-tests/20260730020000_windmill_rls_enable_policy_tables.test.sql
+++ b/packages/data/sql/dbmate/migration-tests/20260730020000_windmill_rls_enable_policy_tables.test.sql
@@ -139,8 +139,30 @@ BEGIN
     CREATE POLICY kbve_windmill_user_all ON windmill.rls_probe_admin_only
         FOR ALL TO windmill_user USING (true) WITH CHECK (true);
 
-    -- 6. reconciliation: once Windmill ships its own windmill_user policy, our
-    --    permissive compat policy must go, or it would OR-widen past theirs.
+    -- 6a. reconciliation, partial coverage: an upstream SELECT-only policy
+    --     must NOT evict the FOR ALL compat policy -- policies apply per
+    --     command, and dropping ours would deny-all every write. Writes must
+    --     keep working and the compat policy must survive.
+    CREATE POLICY see_own_select_upstream ON windmill.rls_probe_admin_only
+        FOR SELECT TO windmill_user USING (workspace_id = 'ws-a');
+    PERFORM windmill.enforce_policy_rls(true);
+    IF NOT EXISTS (
+        SELECT 1 FROM pg_policy p
+        WHERE p.polrelid = 'windmill.rls_probe_admin_only'::regclass
+          AND p.polname = 'kbve_windmill_user_all'
+    ) THEN
+        RAISE EXCEPTION 'fail: SELECT-only upstream policy evicted the FOR ALL compat policy (write paths now deny-all)';
+    END IF;
+
+    SET LOCAL ROLE windmill_user;
+    INSERT INTO windmill.rls_probe_admin_only (workspace_id) VALUES ('ws-write-probe');
+    DELETE FROM windmill.rls_probe_admin_only WHERE workspace_id = 'ws-write-probe';
+    RESET ROLE;
+    DROP POLICY see_own_select_upstream ON windmill.rls_probe_admin_only;
+
+    -- 6b. reconciliation, full coverage: once Windmill covers every command
+    --     for windmill_user, our permissive compat policy must go, or it
+    --     would OR-widen past theirs.
     CREATE POLICY see_own_upstream ON windmill.rls_probe_admin_only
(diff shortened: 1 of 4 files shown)


h0lybyte:
Reconciliation was too coarse: any upstream `windmill_user` policy — even `FOR SELECT` only — evicted the `FOR ALL` compat policy, leaving INSERT/UPDATE/DELETE deny-all on the queue tables while the read probe kept reporting healthy. Reproduced against the kilobase image (`17.4.1.069`) with a live Windmill (`main`, 473 sqlx migrations, prod-parity `search_path=windmill` URL): after a simulated upstream SELECT-only policy plus one `enforce_policy_rls(false)` run, `windmill_user` INSERT on `v2_job_queue` failed with `new row violates row-level security policy` while `rls_user_can_read` and `rls_drift()` both stayed green.

Fixed in 0ac503e7f0: the compat policy is now dropped only when upstream permissive policies cover all of SELECT/INSERT/UPDATE/DELETE for `windmill_user` (`polcmd`-aware, `*` or per-command set); partial coverage keeps the compat policy and raises a WARNING for manual review. Verified live: SELECT-only upstream → compat kept, writes work, warning emitted; `FOR ALL` upstream and 4-per-command upstream → compat dropped. Companion test extended with the partial-coverage case (write probe included).

Also verified end-to-end on the same stack: full dbmate chain applies clean, the five real offenders (`account`, `v2_job_*`) heal exactly as on prod, a Windmill job runs green under the new RLS, and `windmill_user` retains full DML on `v2_job_queue`.

PR #15015, opened 2026-07-31
fix(astro-kbve): anchor palworld marker canvas to leaflet overlay pane

h0lybyte:
Markers lagged/shifted during map scroll — the canvas sat outside leaflet's panes and repainted one rAF behind the tiles' CSS transform. Moved it into the overlay pane and re-anchored via `containerPointToLayerPoint` on each draw: panning now carries markers in the same transform as the tiles (no repaint at all mid-drag), full redraw happens on moveend/zoomend. Tooltip hides on movestart/zoomstart to avoid stale hover. Playwright pan+zoom check: markers stay glued to terrain, no console errors.

### DIFF
--- a/apps/kbve/astro-kbve/src/components/palworld/map/ReactPalworldMap.tsx
+++ b/apps/kbve/astro-kbve/src/components/palworld/map/ReactPalworldMap.tsx
@@ -231,9 +231,21 @@ export default function ReactPalworldMap() {
 					}
 					continue;
 				}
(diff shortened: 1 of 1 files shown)


h0lybyte:
Added the cooldown ring per feedback: defeated boss greys out + amber ring depletes around the marker until respawn; exact time remains in the hover tooltip. Verified via mock relay defeat event.

PR #15321, opened 2026-08-04
chore(rn): bump third-party Expo/RN libraries post-SDK 57

h0lybyte:
Follow-up to #15317 (Expo SDK 56 -> 57). That PR moved everything Expo manages through `bundledNativeModules.json`; this one covers the libraries it does not.

**Supersedes #14722.** That PR bumped `react-native-webgpu` in the root manifest only, which would have left the declarative copy in `apps/kbve/kbve-react-native/package.json` pinned at 0.6.3. Dependabot has this blind spot for every RN dependency in the repo, because the app and lib manifests are not a real pnpm workspace — they are declarative copies that have to be updated by hand.

## Bumped

| package | before | after |
| --- | --- | --- |
| `react-native-webgpu` | 0.6.3 | 0.8.1 |
| `typegpu` | 0.11.8 | 0.11.9 |
| `react-native-url-polyfill` | ^3.0.0 | ^4.0.0 |
| `@hcaptcha/react-native-hcaptcha` | ^4.0.0 | ^4.1.0 |

All three manifests (root, `apps/kbve/kbve-react-native`, `packages/npm/rn`) are updated together.

No source changes were needed:

- `react-native-webgpu` 0.8.1 still exports `Canvas`, `useCanvasRef` and `useDevice`, which is the full extent of what `TypeGpuHost.tsx` imports. Its peer ranges (`react-native >=0.81`, `reanimated >=4.2.1`, `worklets >=0.7.2`) are all satisfied by the SDK 57 stack.
- `react-native-url-polyfill` 4.0.0 is a major, but still exports `setupURLPolyfill`, the only entry point `packages/npm/rn/src/auth/supabase.ts` touches.

## Deliberately not bumped

These have newer majors available, but SDK 57 bundles the current versions and `expo-doctor` validates against them. Moving them independently would put the project out of step with the SDK:

| package | pinned | latest |
| --- | --- | --- |
| `react-native-webview` | 13.16.1 | 14.0.1 |
| `@react-native-async-storage/async-storage` | 2.2.0 | 3.1.1 |
| `react-native-gesture-handler` | ~2.32.0 | 3.1.0 |
| `react-native-safe-area-context` | ~5.7.0 | 5.8.1 |
| `react-native-svg` | 15.15.4 | 15.15.5 |

`@shopify/react-native-skia` stays at 2.6.9, already ahead of the 2.6.2 the SDK bundles. `@expo/vector-icons` (^15.1.1) and `react-native-web` (^0.21.2) are already at latest.

## Verification

- `kbve-react-native:typecheck` — green.
- `rn:lint` — 0 errors (55 pre-existing `no-explicit-any` warnings, untouched).
- `expo export -p ios` — 1736 modules, clean.
- `expo export -p web` — 1895 modules, clean.
- `expo-doctor` — 19/20, the only failure being the pre-existing "no lock file" check (structural: the lockfile lives at the repo root).

Not verified: no on-device render, same as #15317. The app needs a dev build to run at all (local wgpu native module + `expo-dev-client`), and `react-native-webgpu` is the one bump here whose behaviour is native-side, so it is worth exercising on a device before relying on the GPU path.


h0lybyte:
Booted this on an iPhone 17 Pro simulator to verify the stack end to end. Pushed 06d7da8 with the fixes that were needed to get there.

The dev build now **builds, installs, and renders the login screen with no redbox**. Device logs are clean apart from an `AuthApiError: Invalid Refresh Token`, which is just a stale session in the simulator's AsyncStorage from a previous install.

Getting there surfaced three pre-existing breakages in the native path — none caused by the Expo or third-party bumps. The iOS dev build had simply never been exercised:

1. **`kbve_wgpu` didn't compile for `aarch64-apple-ios`.** wgpu 29 changed `SurfaceTargetUnsafe::RawHandle.raw_display_handle` to `Option<RawDisplayHandle>`. Wrapped in both the UiKit and Android arms.

2. **Undefined symbol `_kbve_wgpu_create_game` at link time.** It's gated behind `#[cfg(feature = "bevy")]`, but `KbveWgpuView.swift` and `KbveWgpuView.kt` call it unconditionally, so any non-bevy build fails to link. Added a stub returning null, which both callers already treat as "surface unavailable". Note the bevy-featured build can't be produced at all right now: `isometric-game` pulls `tauri-build 2.6.3`, which panics with `missing cargo:dev instruction, please update tauri to latest`. Worth a separate issue.

3. **App redboxed with `'RNGestureHandlerModule' could not be found`.** Four dependencies `@kbve/rn` uses were absent from the app manifest, so autolinking skipped their pods: `react-native-gesture-handler`, `@shopify/react-native-skia`, `@expo/vector-icons`, `react-native-url-polyfill`. Added and the list sorted.

`react-native-webgpu` 0.8.1 compiles and links as a pod. Its **runtime** GPU path is still unverified — the FX screen sits behind `AuthGate`, so reaching it needs a real sign-in. Also relevant: `build:native:ios` in the app's `project.json` wires up `kbve_wgpu:build:ios`, which omits the `bevy` feature the Swift view needs; with the stub it links, but game mode returns null.

clippy is clean and the crate's tests pass. `cargo fmt --check` still fails on an unrelated import ordering in `lib.rs`'s `android_jni` module, which is pre-existing and left alone.

PR #15373, opened 2026-08-05
fix(kubevirt): point virt-handler's legacy cert manager at a real cert

h0lybyte:
Fixes the virt-handler certificate error **at its source**. No log suppression — an earlier revision of this PR muted the line in `vector-config`, that has been reverted and the vector config is untouched here.

## The mount was never the fix

The operator removes it deliberately. `generateDaemonSet` strips the cert secret at create (v1.7.2:374, v1.9.0:395) and `processCanaryUpgrade` unattaches it whenever the cached DaemonSet lacks it (v1.7.2:200-203, v1.9.0:225-228 — byte-identical). Absence is a **one-way migration marker** for virt-handler's migration-TLS topology: the live DaemonSet records that the install has crossed the transition, and the guard stops any render from regressing it.

That makes the #15325 seed Job actively harmful. Measured on the cluster 2026-08-05:

| time | event |
|---|---|
| 02:00:12 | seed patches `ds/virt-handler` — ControllerRevision 5 carries `kubevirt-virt-handler-certs` |
| ~02:00:4x | operator reverts; next revision reuses the March pod-template hash `5bf7cbbb65`, identical to pre-seed |
| 02:00:42 | pod rolls onto `5bf7cbbb65` without the volume; error resumes |

It rolled a node-level component every Argo sync to force a state KubeVirt treats as regressive. Deleted here, with its SA/Role/RoleBinding (`cr-application` prunes).

## The actual defect is the consumer

virt-handler builds a `FileCertificateManager` on `--client-cert-file` / `--client-key-file` (v1.7.2:186, v1.9.0:194) and starts it **unconditionally** (v1.7.2:442, v1.9.0:510):

```go
app.clientcertmanager = bootstrap.NewFileCertificateManager(app.clientCertFilePath, app.clientKeyFilePath)
...
go app.clientcertmanager.Start()
```

Those flags default to `/etc/virt-handler/clientcertificates/tls.crt` — the path the operator guarantees is absent. That poll loop is what logs once a minute. `migrationOldClientTLSConfig` is still wired from it (v1.9.0:816) despite `--migration-cn-types migration` meaning nothing reads it. The two halves of KubeVirt disagree.

## The fix

Repoint those flags at the **vsock client cert**, which is already mounted at `/etc/virt-handler/vsockclientcertificates` and is the same shape — a virt-handler client cert issued off `kubevirt-ca`. The manager loads a real certificate, the error stops, and no volume is added for the operator to strip. The guard only ever touched the volume; it does not inspect args, so this is stable across reconciles.

`type: json` so the ops **append** to args. A strategic patch would replace the list wholesale and pin operator-managed flags into this file, breaking silently on upgrade.

Verified with `kubectl apply --server-side --dry-run=server` against the live cluster: `serverside-applied`.

## Impact

Rolls virt-handler **once**, on merge. No VMIs running, so node-level VM management pauses briefly and no guest is affected. Unlike the seed, this is one roll, not one per sync.

**Caveat for a second node:** the legacy migration-proxy client path at :313 would then present the vsock cert's CN rather than the dedicated one. Same CA, and that path is only reachable when migrating against a virt-handler too old for `--migration-cn-types`. Single node today, so unreachable — revisit when a second node joins.

## Upgrade

v1.9.0 changes nothing here: guard **and** consumer are both unchanged. kubevirt/kubevirt#16255 and #16264 report this since v1.4. The proper upstream fix is to skip the legacy cert manager when migration CN types are set — worth filing separately.


h0lybyte:
Follow-up: root cause is sharper than the PR body states, and it strengthens the case.

**The guard is not a bug.** Absence of the mount is a deliberate **one-way migration marker** for virt-handler's migration-TLS topology. `generateDaemonSet` strips the secret at create, and the guard then blocks reattachment so an already-migrated install cannot regress to the legacy secret-mounted arrangement partway through a canary rollout. The live DaemonSet *is* the migration-state marker: "no Secret" means "this install has crossed the transition, do not go backward."

That reframes the seed Job. It was not merely losing a race — it was attempting to reverse a transition KubeVirt deliberately treats as one-way, rolling virt-handler on each attempt. Deleting it is correct on those grounds alone.

**The actual bug is on the consumer side**, and it is what emits the log line. virt-handler builds a cert manager on the deliberately-unmounted path and starts it unconditionally — identical in both versions:

```go
// v1.7.2:186 / v1.9.0:194
app.clientcertmanager = bootstrap.NewFileCertificateManager(app.clientCertFilePath, app.clientKeyFilePath)
// v1.7.2:442 / v1.9.0:510
go app.clientcertmanager.Start()
```

`clientCertFilePath` defaults to `/etc/virt-handler/clientcertificates/tls.crt` in both. So the operator guarantees the file is absent and virt-handler polls for it forever. `migrationOldClientTLSConfig` is still wired from that manager at v1.9.0:816 despite `--migration-cn-types migration` meaning nothing reads it.

Net: the two halves of KubeVirt disagree. Nothing is broken on this cluster — it is a pure log artifact — so suppression is the correct resolution rather than a workaround.

This also closes out the upgrade question from both directions: the guard is unchanged in v1.9.0 **and** the consumer is unchanged. v1.9.0 will not silence this. Upgrade for the other five minor versions of fixes, not for this.

PR #15451, opened 2026-08-06
fix(reel): keep live HLS ahead of the player on transcode-heavy sources

h0lybyte:
## Problem

Streaming a 4K hevc source: probe correctly routes `transcode_hls` (full libx264 re-encode), but the encode runs ~0.4× realtime on the 2-CPU pod — made worse by the background library transcode of the same file racing it for the cores. The player burns through produced segments, hits the live edge, exhausts the hls.js fatal-error ladder, and the stream dies ("off air") instead of buffering.

## Fix (four layers)

**Attack the encode rate**
- `REEL_HLS_MAX_HEIGHT` (default 1080, 0=off): HLS transcode/live-encode adds `-vf scale=-2:min(1080\,ih)` — never upscales; a 2160p source encodes ~4× fewer pixels. The library `.reel.mp4` transcode stays full-res.
- Viewer priority: the manifest handler aborts an in-flight background transcode of the same torrent before starting a `TranscodeHls` job, and `should_auto_transcode` skips entries whose HLS job is Starting/Live. No more two ffmpeg encodes splitting the same cores.

**Attack death-instead-of-buffer**
- `REEL_HLS_SEGMENT_WAIT_MS` (default 5000): while HLS is Live, a missing segment waits for the encoder (250ms polls) before 404ing — edge starvation becomes a delayed response, not an error strike.
- Player (reelService.ts): fatal frag/level load failures get their own 40-retry buffering budget (reset on FRAG_BUFFERED) with a 'Buffering — waiting for stream data' status, instead of consuming the 6-strike fatal ladder. 401/manifest errors keep the existing recovery paths.

## Tests
- `cargo test -p reel`: 162 passed (was 158) — scale filter, auto-transcode HLS gate, segment wait/deadline, config defaults
- clippy clean (2 pre-existing warnings untouched)
- `astro check`: 0 errors in media components (26 repo-wide pre-existing)

mdx 0.1.36 → 0.1.37

### DIFF
--- a/apps/kbve/astro-kbve/src/components/media/reelService.ts
+++ b/apps/kbve/astro-kbve/src/components/media/reelService.ts
@@ -688,10 +688,23 @@ export class ReelPlayer {
 			// up once repeated recovery attempts stop working.
 			let netRetries = 0;
 			let mediaRetries = 0;
+			let starveRetries = 0;
 			const MAX_RECOVER = 6;
+			// Live-edge starvation: the encoder simply hasn't produced the next
+			// segment yet, so frag/playlist loads fail while nothing is actually
+			// broken. That's buffering, not an error — wait and reload with a far
+			// larger budget than the real-error ladder.
+			const MAX_STARVE = 40;
+			const STARVATION_DETAILS = new Set<string>([
+				Hls.ErrorDetails.FRAG_LOAD_ERROR,
+				Hls.ErrorDetails.FRAG_LOAD_TIMEOUT,
+				Hls.ErrorDetails.LEVEL_LOAD_ERROR,
+				Hls.ErrorDetails.LEVEL_LOAD_TIMEOUT,
+			]);
 			hls.on(Hls.Events.FRAG_BUFFERED, () => {
 				netRetries = 0;
 				mediaRetries = 0;
+				starveRetries = 0;
 				this.resurrects = 0;
 			});
 			hls.on(Hls.Events.ERROR, (_evt, data) => {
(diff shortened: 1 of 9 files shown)


h0lybyte:
Also bumped reel requests cpu 4→8 / mem 1Gi→2Gi (limit stays 16). Node talos-4bn-whr is 64-core at 67% requests committed — guarantees the live encode ≥8 cores under contention. Note: limit was already 16 with REEL_ENCODE_THREADS=16; the 0.4× realtime came from the double-encode contention + 4K hevc decode cost this PR removes, not from a 2-CPU cap.

h0lybyte:
Third commit closes the last known live-HLS edge case: **dead encoder = frozen playlist hang**.

With the patient player from this PR, a crashed ffmpeg became a silent forever-stall — playlist reloads return 200, no segment 404s, no fatal error. Now:

- **server**: every HLS failure/abort path (and restart reconcile) appends `#EXT-X-ENDLIST` to media playlists (`finalize_event_playlists`, skips master playlists, idempotent) — the stream terminates deterministically
- **player**: stall watchdog escalates to resurrect after 6 consecutive no-progress kicks (~30s) instead of kicking `startLoad` forever; on `ended` the client checks server HLS status — `Failed` → resurrect (restart encode) rather than presenting a mid-movie end as finished

Also verified: the leech→seed completion flip is already handled server-side by `wait_hls_settle` + live-HLS adopt (newer than the ranked-gap list) — non-adopt only happens for failed live jobs, which the resurrect + ENDLIST paths now cover.

163 reel tests; astro check still 0 media errors.

PR #15457, opened 2026-08-06
fix(reel): terminate dead-encoder streams + guarantee encode pod resources

h0lybyte:
## Summary

Follow-up to #15451 — that squash-merge captured the HLS rate fixes and both resource bumps, so after a rebase this PR carries only the **dead-encoder termination** work (plus its release bump):

- **server** (`hls.rs`): a crashed/killed HLS ffmpeg left an EVENT playlist without `#EXT-X-ENDLIST`, so the (now patient) player polled a frozen playlist forever. `finalize_event_playlists` appends the tag to media playlists (skips master, idempotent) on every failure/abort path and on restart reconcile.
- **player** (`reelService.ts`): stall watchdog escalates to a full resurrect after 6 consecutive no-progress kicks (~30s) instead of kicking `startLoad` forever; on `ended` the client checks server HLS status — `Failed` → restart the stream rather than presenting a mid-movie end as finished.
- **mdx** 0.1.37 → 0.1.38 (reel image change → release bump).

## Tests
`cargo test -p reel`: 163 passed (adds finalize idempotency/master-skip test); astro check 0 media errors.

### DIFF
--- a/apps/kbve/astro-kbve/src/components/media/reelService.ts
+++ b/apps/kbve/astro-kbve/src/components/media/reelService.ts
@@ -565,11 +565,14 @@ export class ReelPlayer {
 
 	// A stall watchdog: if the playhead stops advancing while playback is
 	// intended and the buffer is starved (readyState below HAVE_FUTURE_DATA),
-	// kick the active recovery path. Runs until torn down, so a stream that
-	// hangs at the download edge resumes on its own without a manual refresh.
+	// kick the active recovery path. A frozen EVENT playlist (dead encoder)
+	// never errors — playlist reloads keep returning 200 — so repeated
+	// no-progress kicks escalate to a full resurrect instead of kicking forever.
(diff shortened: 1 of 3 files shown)


h0lybyte:
Fixed the conflict: #15451's squash-merge captured the first three commits of this branch (HLS rate fixes + both resource bumps), so this PR was re-carrying already-merged changes and conflicting on `reel.yaml`. Rebased onto dev — the resource commits collapsed away; what remains is exactly the dead-encoder termination work:

- `hls.rs`: `finalize_event_playlists` appends `#EXT-X-ENDLIST` on all HLS failure/abort paths + restart reconcile
- `reelService.ts`: watchdog escalates to resurrect after ~30s of no progress; `ended` while server reports HLS Failed → restart instead of fake movie-end
- mdx 0.1.37 → 0.1.38 (was missing — reel image change needs the release bump)

163 reel tests pass.

PR #15469, opened 2026-08-06
fix(vector): end-to-end acks on ClickHouse sinks (keep memory buffers)

### LINKED ISSUE #13099: [HIGH] Vector ClickHouse sinks use in-memory buffers
**Item:** 🟠 Reliability of #12973

`vector/manifests/vector-config.yaml:446-469` (`buffer: type: memory`): all buffered logs/alerts lost on Vector restart/OOM (4Gi limit).

**Fix:** `buffer: type: disk` on the mounted `/vector-data-dir`.

Split out from #12973 (apps/kube infra refactor audit).

h0lybyte:
## What

Addresses #13099 without switching to disk buffers — decision: memory buffers stay, no continuous write I/O from Vector's own buffering.

Both `clickhouse` and `clickhouse_alerts` sinks gain:

```yaml
acknowledgements:
  enabled: true
```

## Why this closes the loss gap

- `kubernetes_logs` only advances its checkpoint after ClickHouse acks the batch → on Vector restart/OOM, unacked lines are re-read from the kubelet's pod log files (already on disk regardless) — zero added write amplification, no loss across restarts.
- `alertmanager_webhook` (http_server) holds the HTTP response until delivery; failures return non-200 → Alertmanager retries per its own retry loop.
- Memory buffer `when_full: block` (default) backpressures during a ClickHouse outage; log files persist meanwhile.

Residual (accepted): lines lost only if the pod log file rotates away during an outage longer than kubelet retention.

Config-version bumped `v10-tracing-level-fix` → `v11-sink-e2e-acks`. Embedded `vector.yml` decoded + re-parsed to verify only the two ack blocks changed.

Closes #13099

Docs: [kbve.com/application/kubernetes](https://kbve.com/application/kubernetes/)

### DIFF
--- a/apps/kube/vector/manifests/vector-config.yaml
+++ b/apps/kube/vector/manifests/vector-config.yaml
@@ -5,14 +5,15 @@ metadata:
     namespace: vector
     annotations:
         kbve.sh/last-updated: '2026-08-06T00:00:00Z'
-        kbve.sh/config-version: v11-sink-e2e-acks
+        kbve.sh/config-version: v12-alerts-disk-buffer
     labels:
         kbve.com/category: observability
         kbve.com/stack: core
         kbve.com/managed-by: argocd
 data:
     vector.yml:
-        "api:\n  enabled: true\n  address: 0.0.0.0:9001\n\nsources:\n  kubernetes_logs:\n\
+        "api:\n  enabled: true\n  address: 0.0.0.0:9001\n\ndata_dir: \"/vector-data-dir\"\n\n\
+        sources:\n  kubernetes_logs:\n\
         \    type: kubernetes_logs\n    self_node_name: \"${VECTOR_SELF_NODE_NAME}\"\n\
         \    exclude_paths_glob_patterns:\n      - \"**/vector_*/**\"\n      - \"**/kube-system_*/**\"\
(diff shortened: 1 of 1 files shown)


h0lybyte:
Added per follow-up: `clickhouse_alerts` now uses a 256MiB disk buffer (Vector's minimum) on the hostPath-backed `/vector-data-dir` — alerts lack upstream replay, and their volume makes the I/O negligible. Logs sink unchanged (memory + acks, kubelet files as replay source). Global `data_dir` now explicit so buffers land on the mounted volume. Config-version `v12-alerts-disk-buffer`.

h0lybyte:
Correction, caught by running `vector validate` against the real 0.54 binary: **`kubernetes_logs` does not support end-to-end acknowledgements in Vector 0.54** — the acks flag on the logs sink was a no-op and Vector warns "Silent data loss could occur." Removed it from the `clickhouse` logs sink (`e914d9c`).

Net behavior for logs: bounded memory buffer (10k events), drained on graceful shutdown; loss only on hard crash/OOM, capped at buffer size. That's the memory-buffer tradeoff as originally chosen — just without the false ack safety net in the description above. Alerts path unchanged: http_server supports acks, disk buffer holds through restarts.

PR #15493, opened 2026-08-07
feat(stalwart): MTA-only mail server for $username@herbmail.com

h0lybyte:
## Summary

Stands up [Stalwart](https://github.com/stalwartlabs/stalwart) as an **MTA-only** mail server for `$username@herbmail.com`. Stalwart terminates SMTP (shared LB IP, ports 25/465/587), runs SPF/DKIM/spam checks, then fires a data-stage [MTA hook](https://stalw.art/docs/api/mta-hooks/overview/) at axum-herbmail, which resolves recipients against `profile.username` and inserts into `mail.messages` — answering `{"action":"discard"}` so Stalwart never stores mail. Single copy, RLS per user, 90-day pg_cron purge. Internal user-to-user messages insert directly (`via='internal'`) and skip SMTP.

Docs: [apps/kube/stalwart/README.md](https://github.com/KBVE/kbve/blob/trunk/stalwart-mail/apps/kube/stalwart/README.md) — wizard, hook, DNS/PTR setup.

## Changes

- **apps/kube/stalwart/**: Deployment (`stalwartlabs/stalwart:v0.16.16`, NET_BIND_SERVICE only), 8Gi longhorn PVC, LB service sharing `142.132.206.71` (`sharing-key public-71`), ClusterIP for admin/submission, cert-manager Certificate for `mail.herbmail.com`; registered in app-of-apps.
- **kbve-gateway + arpg LB**: add `stalwart` to `lbipam.cilium.io/sharing-cross-namespace` (sharing is a mutual allowlist).
- **dbmate**:
  - `20260807120000_stalwart_role_schema.sql` — scoped `stalwart` role + schema (forgejo pattern), `search_path` pinned so Stalwart's auto-managed tables land in its own schema.
  - `20260807121000_mail_schema_init.sql` — `mail.messages` (+ indexes, RLS select-own), `public.stalwart_ingest` SECURITY DEFINER owned by postgres, EXECUTE for service_role only, pg_cron 90d purge (skips gracefully when pg_cron absent locally).
- **axum-herbmail**: `POST /hooks/stalwart` — Bearer auth against `STALWART_HOOK_SECRET` (503 fail-closed when unset), own 32MB body limit outside the global 1MB cap, PostgREST RPC ingest, `discard` on accept / `reject 550` when no recipient resolves. Deployment env reads the (optional) `stalwart-hook-secret` sealed secret; `seal-stalwart-hook-secret.sh` generates it.

## Post-merge (manual)

1. `ALTER ROLE stalwart WITH PASSWORD ...` after migration deploy.
2. Stalwart setup wizard (PostgreSQL store → kilobase-rw, db supabase, user stalwart) + MTA hook config with the sealed bearer secret.
3. DNS: A/MX/SPF/DKIM/DMARC + provider PTR for `mail.herbmail.com` → `142.132.206.71`.

## Caveats

- Shared LB IP with default `externalTrafficPolicy: Cluster` may SNAT client IPs — weakens SPF/DNSBL; README documents the Local-policy / dedicated-IP options.
- Outbound send path (agent replies via submission) is a follow-up; keep accounts reply-only + rate-limited before opening it.


h0lybyte:
Subagent audit results — fixed in bcd2ff8:

**Blocker (fixed)**: UID 2000 could not bind ports 25/443/465 under kubelet — `capabilities.add: NET_BIND_SERVICE` never reaches the ambient set for non-root, and the local Docker lab masked it (`ip_unprivileged_port_start=0` is a Docker default, not a kubelet one). Fixed with the pod-level safe sysctl.

**Should-fix (fixed)**:
- Public pre-auth 32MB body parse on `/hooks/stalwart` (reachable via herbmail.com HTTPRoute): Bearer check moved into router middleware (rejects on headers, before body buffering) + 30s TimeoutLayer restored (bypassing the global stack had also dropped its timeout).
- `/etc/stalwart` writability: read-only subPath mount diverged from the verified lab; now emptyDir seeded by initContainer.
- RLS on `profile.username` could have made `stalwart_ingest` silently resolve zero recipients (550 on all inbound mail): defensive GRANT + SELECT policy for postgres, mirroring the custom-access-token-hook migration. Plus NULL-array guard.
- PVC 8Gi → 2Gi (state lives in postgres; longhorn overcommitted).

**Deploy-order note (not a code change)**: the hook endpoint ships in the next herbmail image release — until `ghcr.io/kbve/herbmail` is rebuilt past 0.1.2 and rolled, Stalwart hook calls 404 → 451 tempfail (senders retry, no mail lost). Configure the Stalwart MTA hook after the herbmail image rolls.

Verified clean by the audit: LB sharing mutual allowlist + port disjointness on 142.132.206.71, PostgREST exposure (public schema exposed, RPC service_role-only), SQL injection surface, empty search_path qualification, seal scripts, cert-manager HTTP-01 path via gateway, no schema/port collisions. `cargo test`: 8/8.

